PR #1, opened 2022-07-07
melpomene: make `tracing` feature flags additive

hawkw:
Currently, the `tracing` behavior in Melpomene is controlled by a pair
of feature flags, "trace-modality" and "trace-fmt". These select between
emitting traces to stdout using `tracing_subscriber::fmt`, and emitting
to Modality using the `tracing-modality` crate. However, these feature
flags are not additive. If they are both enabled, the code will not
compile, since there are two implementations of the `setup_tracing`
function, which are conditionally compiled based on whether those
feature flags are enabled. If both are enabled, both functions will be
present, resulting in a compiler error.

This commit changes this code so that there is a single implementation
of the `setup_tracing` function, which will construct a `tracing`
subscriber that includes either `fmt`, Modality, or _both_, depending on
the feature combination. Now, Melpomene should build even with
`--all-features` or similar.

Now that the `fmt` layer is built manually rather than using
`fmt::init()` "easy mode" helper, I've also changed the filter
environment variable from "RUST_LOG" to "MELPOMENE_TRACE". I thought
this was nicer, but we can continue using the default env var if that's
preferable.

### DIFF
--- a/source/melpomene/src/sim_tracing.rs
+++ b/source/melpomene/src/sim_tracing.rs
@@ -1,9 +1,21 @@
-#[cfg(feature = "trace-fmt")]
-pub fn setup_tracing() {
-    tracing_subscriber::fmt::init();
-}
+const ENV_FILTER: &str = "MELPOMENE_TRACE";
 
-#[cfg(feature = "trace-modality")]
 pub fn setup_tracing() {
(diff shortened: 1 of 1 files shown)


jamesmunns:
Tested this works with both subscribers! Thanks :)

PR #2, opened 2022-07-07
melpomene: tracing improvements

hawkw:
(This includes changes from #1 because that PR was opened from a fork
and I can't nicely stack PRs when one branch is on a fork)

This branch makes several small improvements to `tracing` in Melpomene:

 * melpomene: make `tracing` feature flags additive
 
   Currently, the `tracing` behavior in Melpomene is controlled by a
   pair of feature flags, "trace-modality" and "trace-fmt". These select
   between emitting traces to stdout using `tracing_subscriber::fmt`,
   and emitting to Modality using the `tracing-modality` crate. However,
   these feature flags are not additive. If they are both enabled, the
   code will not compile, since there are two implementations of the
   `setup_tracing` function, which are conditionally compiled based on
   whether those feature flags are enabled. If both are enabled, both
   functions will be present, resulting in a compiler error.
   
   This commit changes this code so that there is a single
   implementation of the `setup_tracing` function, which will construct
   a `tracing` subscriber that includes either `fmt`, Modality, or
   _both_, depending on the feature combination. Now, Melpomene should
   build even with `--all-features` or similar.
   
   Now that the `fmt` layer is built manually rather than using
   `fmt::init()` "easy mode" helper, I've also changed the filter
   environment variable from "RUST_LOG" to "MELPOMENE_TRACE". I thought
   this was nicer, but we can continue using the default env var if
   that's preferable.

 * melpomene: minor improvements to tracing configuration
 
 * melpomene: add some spans to the kernel/userspace threads
 
 * kernel: change the "initialized heap" event to format the addr in hex

Closes #1

### DIFF
--- a/source/Cargo.lock
+++ b/source/Cargo.lock
@@ -36,6 +36,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "atty"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
+dependencies = [
+ "hermit-abi",
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


jamesmunns:
Totally works for me! Thanks for the cleanup :)

PR #3, opened 2022-07-07
Basic working multi-driver demo using async bbq

jamesmunns:
This implements a quad-waitcell version of bbq, as well as a basic PoC with two sort-of drivers. This data structure is intended for drivers such as a serial port.


source/kernel/src/bbq.rs
```diff
@@ -84,14 +103,166 @@ impl Drop for BBQStorage {
...
+                Ok(wgr) => {
+                    trace!(
+                        size = wgr.len(),
+                        max = max,
+                        "Got bbqueue max write grant",
+                    );
+                    let waker = unsafe {
+                        let ptr = addr_of_mut!((*self.our_waitcells.as_ptr()).commit_waitcell);
+                        NonNull::new_unchecked(ptr)
+                    };
```
hawkw, line 193: whew, you love 2 see it

source/kernel/src/bbq.rs
```diff
@@ -57,6 +72,8 @@ pub async fn new_bidi_channel(
         BBQBidiHandle {
             producer: a_prod,
             consumer: b_cons,
+            our_waitcells: NonNull::new_unchecked(&storage.a_wait as *const BBQWaitCells as *mut BBQWaitCells),
+            other_waitcells: NonNull::new_unchecked(&storage.b_wait as *const BBQWaitCells as *mut BBQWaitCells),
```
hawkw, line 76: can't this just be the (safe) `From<&'_ T> for NonNull` impl?
```suggestion
            our_waitcells: NonNull::from(&storage.a_wait),
            other_waitcells: NonNull::from(&storage.b_wait),
```

source/kernel/src/bbq.rs
```diff
@@ -69,6 +86,8 @@ pub async fn new_bidi_channel(
         BBQBidiHandle {
             producer: b_prod,
             consumer: a_cons,
+            our_waitcells: NonNull::new_unchecked(&storage.b_wait as *const BBQWaitCells as *mut BBQWaitCells),
+            other_waitcells: NonNull::new_unchecked(&storage.a_wait as *const BBQWaitCells as *mut BBQWaitCells),
```
hawkw, line 90: same here:
```suggestion
            our_waitcells: NonNull::new_unchecked(&storage.b_wait),
            other_waitcells: NonNull::new_unchecked(&storage.a_wait),
```

source/kernel/src/bbq.rs
```diff
@@ -84,14 +103,166 @@ impl Drop for BBQStorage {
...
+use abi::bbqueue_ipc::{
+    GrantR as InnerGrantR,
+    GrantW as InnerGrantW,
+};
+
+pub struct GrantW {
+    grant: InnerGrantW<'static>,
+    // When we commit, we trigger OUR commit waitcell
+    our_commit: NonNull<WaitCell>,
+}
```
hawkw, line 115: these `Grant` types need to keep the `BBQStorage`'s ref count alive, right? how is that ensured exactly?

source/melpomene/src/main.rs
```diff
@@ -69,27 +73,79 @@ fn kernel_entry() {
...
+                    tracing::warn!("Driver A: Reading...");
+                    let rgr = a_ring.read_grant().await;
+                    tracing::warn!(
+                        buf = ?rgr.deref(),
+                        "Driver A: Got data"
+                    );
+                    let len = rgr.len();
+                    rgr.release(len);
+                }
+            };
```
hawkw, line 113: consider sticking a span on these?
```suggestion
            }.instrument(tracing::info_span!("Driver A");
```
(of course, we'd have to add `use tracing::Instrument;` somewhere)

source/melpomene/src/main.rs
```diff
@@ -69,27 +73,79 @@ fn kernel_entry() {
...
+                    tracing::warn!("Driver B: Reading...");
+                    let rgr = b_ring.read_grant().await;
+                    tracing::warn!(
+                        buf = ?rgr.deref(),
+                        "Driver B: Got data"
+                    );
+                    let len = rgr.len();
+                    rgr.release(len);
+                }
+            };
```
hawkw, line 144:
```suggestion
            }.instrument(tracing::info_span!("Driver B"));
```

hawkw (approved):
this is awesome!

i had some tiny unimportant suggestions, take them or leave them

source/kernel/src/bbq.rs
jamesmunns: It's not! That's a good call. In the "real" version of bbqueue, grants inherit the lifetime param from their producer/consumer, which derive it from the lifetime of the backing storage, which 99% of the time is `'static`. When I hacked this all up for the mnemos version of bbqueue, that kind of went out the window.

I *think* I want to kick this can down the road, but I would like to open an issue to make sure this gets fixed. I'm not sure whether I want the grants holding an actual heap handle, or add a manual incr/decr refcount on an arc.

source/kernel/src/bbq.rs
jamesmunns: So uh, TIL about `From<&'_ T> for NonNull`.

source/kernel/src/bbq.rs
jamesmunns: Ehhh, just changed my mind, I'll go ahead and add the Arc here, it should just be a pointer.

source/kernel/src/bbq.rs
jamesmunns: Fixed this in [31d9370](https://github.com/tosc-rs/mnemos/pull/3/commits/31d93705ffd13380b0f1be85ef7f3061ba594fad)

source/kernel/src/bbq.rs
```diff
@@ -84,14 +106,167 @@ impl Drop for BBQStorage {
...
+                Err(_) => {
+                    trace!("awaiting bbqueue exact write grant");
+                    // Uh oh! Couldn't get a send grant. We need to wait for the OTHER reader
+                    // to release some bytes first.
+                    match self.side {
+                        Side::ASide => &self.storage.b_wait,
+                        Side::BSide => &self.storage.a_wait,
+                    }.release_waitcell.wait().await.unwrap();
+                    trace!("awoke for bbqueue exact write grant");
+                },
```
hawkw, line 240:
```suggestion
                    trace!(
                        size = size,
                        side = ?self.side,
                        "Got bbqueue exact write grant",
                    );
                    return GrantW {
                        grant: wgr,
                        side: self.side,
                        storage: self.storage.clone(),
                    }
                },
                Err(_) => {
                    trace!(side = ?self.side, size, "awaiting bbqueue exact write grant");
                    // Uh oh! Couldn't get a send grant. We need to wait for the OTHER reader
                    // to release some bytes first.
                    match self.side {
                        Side::ASide => &self.storage.b_wait,
                        Side::BSide => &self.storage.a_wait,
                    }.release_waitcell.wait().await.unwrap();
                    trace!(side = ?self.side, size, "awoke for bbqueue exact write grant");
                },
```

source/kernel/src/bbq.rs
```diff
@@ -84,14 +106,167 @@ impl Drop for BBQStorage {
...
+            }
+        }
+    }
+
+    pub async fn send_grant_exact(&self, size: usize) -> GrantW {
+        loop {
+            match self.producer.grant_exact(size) {
+                Ok(wgr) => {
+                    trace!(
+                        size = size,
```
hawkw, line 222: tiny nit: this could just be
```suggestion
                        size,
```

source/kernel/src/bbq.rs
```diff
@@ -84,14 +106,167 @@ impl Drop for BBQStorage {
...
+impl BBQBidiHandle {
+    // async fn send_grant(buf_len: usize) -> GrantW
+    // async fn read_grant() -> GrantR
+    pub async fn send_grant_max(&self, max: usize) -> GrantW {
+        loop {
+            match self.producer.grant_max_remaining(max) {
+                Ok(wgr) => {
+                    trace!(
+                        size = wgr.len(),
+                        max = max,
```
hawkw, line 193: nit: could just be
```suggestion
                        max,
```

PR #7, opened 2022-07-08
fix a whole bunch of annoying clippy lints

hawkw:
I have clippy enabled by default in VS Code, and it currently complains
about a bunch of stuff...most of which is just kind of annoying
pedantry. This branch fixes all those lints; in some cases, this
actually does make the code somewhat nicer (IMO), but mostly it makes
the warnings in my VS Code "problems" window go away.

I've applied the fix for each lint in a separate commit. If there are
any of these changes we would prefer not to apply, the individual commit
can be backed out and replaced with an `#[allow(...)]` attribute or
whatever, as appropriate.

Hope this isn't too annoying of a change...I can also just turn off clippy for
this repo, or learn to live with the warnings.


jamesmunns:
Love it!

PR #8, opened 2022-07-08
melpo: use Tokio to run simulator services (sleeps)

hawkw:
This builds on PR #5, but instead of spawning a separate OS thread for
every `Sleepy` future, we instead change the simulator to run in a
single-threaded Tokio runtime, with each sleep driven by a task spawned
on that runtime. This way, we only need a single thread to drive all of
the simulator services.

In the future, we can also use the same runtime to drive any simulator
services that use I/O (e.g. implementing a simulated serial port using a
host TCP socket). We may also want to use Tokio synchronization
primitives to block the simulator rt on kernel initialization, instead
of the current spin loop, but I'll address that in a separate PR.

### DIFF
--- a/source/melpomene/Cargo.toml
+++ b/source/melpomene/Cargo.toml
@@ -3,7 +3,12 @@ name = "melpomene"
 version = "0.1.0"
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+# See more keys and their definitions at
+# https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies.tokio]
+version = "1.19"
+features = ["rt", "time", "macros"]
 
 [dependencies.tracing]
 version = "0.1.35"
--- a/source/melpomene/Cargo.toml
+++ b/source/melpomene/Cargo.toml
@@ -3,8 +3,7 @@ name = "melpomene"
 version = "0.1.0"
 edition = "2021"
 
-# See more keys and their definitions at
-# https://doc.rust-lang.org/cargo/reference/manifest.html
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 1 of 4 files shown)


source/melpomene/Cargo.toml
```diff
@@ -3,7 +3,12 @@ name = "melpomene"
 version = "0.1.0"
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+# See more keys and their definitions at
+# https://doc.rust-lang.org/cargo/reference/manifest.html
```
hawkw, line 7: ugh didn't mean to reformat this
```suggestion
# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```

jamesmunns:
As of now, this causes the `tracing-modality` library to panic:

```
     Running `/home/james/tosc/mnemos/source/target/debug/melpomene`
   0.026588434s TRACE Melpo: mio::poll: registering event source with poller: token=Token(0), interests=READABLE | WRITABLE
========================================
   0.027037127s  INFO Melpo: melpomene: Kernel started.
thread 'main' panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.', /home/james/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.19.2/src/runtime/enter.rs:39:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
   0.027277799s TRACE Melpo: mio::poll: deregistering event source from poller
   0.051888647s  INFO Kernel: kernel: Initializing heap start=0x7f396400ffd0 size=196608
```

Following up with folks on the Modality team to see what the best approach for getting this sorted is.

jamesmunns:
Now that #16 has landed, we can merge this. Thanks @hawkw !

PR #12, opened 2022-07-11
Add a simple TCP loopback

jamesmunns:
Hey it's james, back with another terrible thread solution to a simulation problem.

This time, I use an OS thread to take ownership of the bidi buffer and directly feed bytes into and out of a TCP port, mapped to port 9999.

This can be opened locally with ncat to see the loopback:

```shell
stty -icanon -echo && ncat 127.0.0.1 9999
hello world!
```

### DIFF
--- a/source/kernel/src/bbq.rs
+++ b/source/kernel/src/bbq.rs
@@ -200,9 +200,8 @@ impl GrantR {
     }
 }
 
+// async methods
 impl BBQBidiHandle {
-    // async fn send_grant(buf_len: usize) -> GrantW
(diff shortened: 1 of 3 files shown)


jamesmunns:
This also comes with a loopback driver, though I'll likely be replacing this shortly with the older "crowtty" driver implementation, which provides framing and virtual ports over the single physical port.

jamesmunns:
@hawkw this can likely be done much better with tokio once the modality issues are resolved.

hawkw (commented):
nice!

PR #15, opened 2022-07-12
Serial Multiplexing

jamesmunns:
This PR adds the "serial mux" which allows for virtual serial ports over a single physical serial link (or in the current case, a single simulated serial port over a TCP stream).

This also significantly reworks the async BBQueue wrapper to allow for three flavors:

* bidi - a bidirectional, dual-spsc channel
* mpsc - a multiple (async mutex'd) producer, single consumer one-way channel
* spsc - a single producer, single consumer one-way channel.


jamesmunns:
This change is now complete. You can reproduce it using the following steps. There is also a [demo video on twitter](https://twitter.com/bitshiftmask/status/1547332011500736515).

## Console one

Shows tracing output. Opens port 9999 as the multiplexed port

```sh
cd source/melpomene
MELPOMENE_TRACE=trace cargo run
```

## Console two

Run the multiplexor server. Connects to port 9999, and opens port 10000 (mapped to port 0), and port 10001 (mapped to port 1).

```sh
cd tools/crowtty
cargo run
```

## Console three

This console watches the 1/sec "hello" output hosted on port 1.

```sh
stty -icanon -echo && ncat 127.0.0.1 10001
# OR
telnet 127.0.0.1 10001
```

## Console four

This console connects to the loopback service on port 0.

```sh
# this doesn't line buffer, so the feedback is immediate
stty -icanon -echo && ncat 127.0.0.1 10000
# OR
# this DOES buffer lines, so you'll see line-by-line echos.
# telnet automatically echos while you are typing
telnet 127.0.0.1 10000
```

PR #18, opened 2022-07-15
chore: put everything in one big Cargo workspace

hawkw:
Currently, this repo contains three separate Cargo workspaces: one for
`source/`, which contains the actual mnemOS crates, and then a separate
root workspace for each crate in `tools/`.

This is a bit unfortunate, as it means that it's not possible to run
`cargo run --bin WHATEVER` from the root directory of the repo. Instead,
in order to run `melpomene`, it's necessary to `cd` into `source`.
Similarly, if I want to run tools like `crowtty`, I need to `cd` into
`tools/crowtty`. This is very mildly annoying. The lack of a global
workspace also means that it's not currently possible to run cargo
commands like `cargo build --workspace` and build everything in one go.

This branch rectifies this by moving the workspace's `Cargo.toml` to the
root of the repo, and including all crates in the workspace (both
`source/` and `tools`/). The potential downside of this is that it means
there's a lockfile, `Cargo.toml`, and `.cargo/` directory in the root of
the workspace, which previously was nice and clean, and didn't have a
bunch of random config files and stuff in it. But, I think this is
probably not *that* bad?

If you'd really prefer not to make this change, that's fine --- having
to change dirs was really only a minor inconvenience.


jamesmunns (approved):
My only concert with this is that I have some memory of workspaces for multiple targets not working right/being a pain in the ass.

I can't remember if it was cargo or rust-analyzer's fault, but the original division of workspaces was made to avoid "embedded" crates and "desktop crates" being in the same workspace.

That being said - in this consideration melpo should probably be in the "tools" folder.

I'm happy to merge this now, but it might make sense to back out later when we get closer to working with real hardware.

hawkw:
> My only concert with this is that I have some memory of workspaces for multiple targets not working right/being a pain in the ass.
> 
> I can't remember if it was cargo or rust-analyzer's fault, but the original division of workspaces was made to avoid "embedded" crates and "desktop crates" being in the same workspace.
> 
> That being said - in this consideration melpo should probably be in the "tools" folder.
> 
> I'm happy to merge this now, but it might make sense to back out later when we get closer to working with real hardware.

Ah, hmm, I was wondering if there was some issue like that. I've not had problems with having all the `mycelium` stuff in the same workspace even though it will do builds for both the host target (build tools) and the actual OS target (which is still x86 in my case but uses a very different target.json). But, it's possible that only works because of some specific aspect of the way I've set things up.

If this causes problems, we can definitely undo it later.

PR #19, opened 2022-07-15
chore: add Cargo alias for running Melpomene

hawkw:
This branch adds `cargo melpomene` and `cargo melpo` aliases for running
Melpomene that are equivalent to `cargo run --bin melpomene -- <ARGS>`
but with slightly fewer characters.

This is very minor but I thought maybe nice to have.

Depends on #18


jamesmunns (approved):
I don't have enough cargo aliases in my life :D

PR #20, opened 2022-07-15
refac(melpomene): run TCP serial driver on Tokio

hawkw:
This branch rewrites the TCP serial port simulated driver to run in
Melpomene's Tokio runtime, rather than spawning its own thread. This
should make Melpomene a bit more efficient, since the runtime thread can
run other simulated drivers while the serial driver task is waiting for
IO, instead of doing a bunch of little sleeps.

I *think* this shouldn't result in any real behavioral change, although
I haven't tested it super extensively. If I open `crowtty` and then kill
it after a while, Melpomene still sees the TCP stream
disconnection...but I wasn't sure how I was supposed to actually start
the simulated serial port after launching `crowtty`? Any suggestions on
further testing would be appreciated! :)

### DIFF
--- a/source/melpomene/src/main.rs
+++ b/source/melpomene/src/main.rs
@@ -95,7 +95,7 @@ fn kernel_entry() {
                 // Create the buffer, and spawn the worker task, giving it one of the
                 // queue handles
                 let (mux_ring, tcp_ring) = new_bidi_channel(k.heap(), 4096, 4096).await;
-                spawn_tcp_serial(tcp_ring);
+                spawn_tcp_serial(tcp_ring).await;
 
                 // Now, right now this is a little awkward, but what I'm doing here is spawning
                 // a new virtual mux, and configuring it with:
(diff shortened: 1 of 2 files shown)


jamesmunns:
Something is causing the TCP task to be polled constantly, I sent you a message in matrix with a screencap video

hawkw:
> Something is causing the TCP task to be polled constantly, I sent you a message in matrix with a screencap video

Yup, just fixed that --- that's my fault for copying the blocking code too religiously.  6402f285e4b05e65ec349e8fd33d510d9779afdc fixes that.

I also figured out how to use `crowtty` _immediately_ after opening the PR, so I noticed the issue as soon as I ran with `trace` logging.

jamesmunns:
Awesome, thanks! Can confirm it looks good to me.

Also, if you want to break this into two tasks, that's totally fine with me. The BidiHandles have a `.split()` method that will give you the producer and consumer halves separately to give to each task.

PR #21, opened 2022-07-15
feat(melpo): add a simple CLI for configuring the simulator 

hawkw:
This branch adds a simple CLI for configuring Melpomene, using `clap`.
Currently, the CLI can be used for configuring options for various
`tracing` backends, and for setting the address on which to bind the TCP
socket for the simulated serial port. More options will probably be
added in the future for configuring other simulator parameters and
debugging features.

Check out the new help output:
```
:; cargo melpo --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/melpomene --help`
melpomene 0.1.0

USAGE:
    melpomene [OPTIONS]

OPTIONS:
    -h, --help
            Print help information

        --serial-addr <SERIAL_ADDR>
            Address to bind the TCP listener for the simulated serial port

            [default: 127.0.0.1:9999]

    -V, --version
            Print version information

TRACING OPTIONS:
        --trace <ENV_FILTER>
            Trace filter for `tracing-subscriber::fmt`.

            This requires that Melpomene be built with the "trace-fmt" feature flag enabled.

            [env: MELPOMENE_TRACE=]
            [default: info]

TRACING OPTIONS (TOKIO-CONSOLE):
        --console-addr <ADDR>
            Address to bind the `tokio-console` listener on.

            This requires that Melpomene be built with the "trace-console" feature flag enabled.

            [env: TOKIO_CONSOLE_BIND=]
            [default: 127.0.0.1:6669]

        --console-publish-interval <PUBLISH_INTERVAL>
            The interval between publishing updates to connected `tokio-console` clients.

            This requires that Melpomene be built with the "trace-console" feature flag enabled.

            [env: TOKIO_CONSOLE_PUBLISH_INTERVAL=]
            [default: 1s]

        --console-record-path <RECORD_PATH>
            A file path to save a `tokio-console` recording to.

            If a value is present, a recording will be output to that file. Otherwise, no recording
            will be saved.

            This requires that Melpomene be built with the "trace-console" feature flag enabled.

            [env: TOKIO_CONSOLE_RECORD_PATH=]

        --console-retention <RETENTION>
            How long to retain `tokio-console` data for completed tasks.

            This requires that Melpomene be built with the "trace-console" feature flag enabled.

            [env: TOKIO_CONSOLE_RETENTION=]
            [default: 1h]
```

PR #24, opened 2022-07-16
Add docs

jamesmunns:
Add top level README docs for most main folders.

### DIFF
--- a/source/melpomene/README.md
+++ b/source/melpomene/README.md
@@ -8,7 +8,79 @@ Drivers to implement platform-specific behavior, such as serial ports, are provi
 
 ## Running the simulator
 
-The simulator can be run using `cargo run` from this folder.
+### Using an alias
+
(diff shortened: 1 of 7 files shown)


source/melpomene/README.md
```diff
@@ -0,0 +1,28 @@
+# Melpomene
+
+Melpomene is a desktop simulator for the MnemOS kernel.
+
+Rather than emulating any specific hardware, it runs the kernel in one thread, and userspace in a second thread.
+
+Drivers to implement platform-specific behavior, such as serial ports, are provided.
+
+## Running the simulator
```
hawkw, line 9: maybe worth mentioning that there's also a `cargo melpomene` alias, and that you can `cargo melpomene --help` to get a complete list of CLI options?

hawkw (approved):

source/melpomene/README.md
jamesmunns: Ah, I meant to do this! I even merged #21 so it made sense, then I totally forgot to do that. Will add.

PR #25, opened 2022-07-18
RFC: Driver Registry

jamesmunns:
Work in progress

### DIFF
--- a/rfcs/0000-driver-registry.md
+++ b/rfcs/0000-driver-registry.md
@@ -1,0 +1,143 @@
+# RFC - Drivers and Discovery
+
+## Goal
+
+MnemOS should have some way of "automatically discovering" drivers. This would be used by three main "consumers":
+
+* The kernel itself, which may need to use drivers that are provided at initialization time
+* Drivers, which may depend on other drivers (such as a virtual serial mux driver, which depends on a "physical serial port" driver)
+* Userspace, which will typically interact with drivers through either the stdlib, or other driver crates.
+
+## Background
+
+Right now, when spawning drivers, the process goes roughly as follows:
+
+* The `init()` code manually constructs drivers, and manually creates any shared resources necessary between drivers
+* The `init()` code spawns the driver task(s) onto the kernel executor
+
+This is a little manual, and also won't really work for userspace.
+
+There are also some half baked ideas in the `abi`/`mstd` crates about having a single "message kind" that is an enum of all possible driver types, but this isn't very extensible
+
+## The proposal
+
+This change proposes inntroducing a **registry** of drivers, maintained by the kernel.
+
+* This **registry** would be based on a Key:Value type arrangement:
+    * The **Key** would be a tuple of:
+        * A pre-defined **UUID**, that uniquely defines a "service" that the driver fulfills, e.g. the ability to process a given command type
+        * The `TypeId` of that message
+    * The **Value** would be contain two things:
+        * a `KProducer<T>`, where `type_id::<T>()` matches the `TypeId` in the key
+        * Optionally, if the driver should process messages from userspace, it should provide a function with the signature `fn(&[u8]) -> Result<T>`, where `type_id::<T>()` matches the `TypeId` in the key
+* Drivers would be responsible for registering themselves at `init()` time
+* All `(Uuid, TypeId)` keys must be unique.
+* The registry would have the following APIs:
+    * `kernel.registry().set::<T>(&Uuid, &KProducer<T>) -> Result<()>`
+    * `kernel.registry().get::<T>(&Uuid) -> Option<KProducer<T>>`
+        * This should have a blocking version (that returns an Option)
+        * And also an async version (that just returns a KProducer)
+
+## Details
+
+Some details about why parts of this proposal was chosen:
+
+### Using `Uuid`s to identify drivers
+
+There are a lot of different ways to discover drivers used in other operating systems, including using some kind of path, or a type erased file interface.
+
+This proposal takes an approach somewhat similar to Bluetooth's characteristics:
+
+* A UUID is chosen ahead of time to uniquely identify a service
+* Some "official" UUIDs are reserved for MnemOS and built-in drivers
+* A UUID represents a single service interface
+    * Although any driver could implement the UUID, all implementors are expected to have the same interface
+    * The UUID has no concept of versioning. If a breaking change is needed, a new UUID should be used
+
+In practice, any unique/opaque identifier could be used, but a UUID is used for familiarity.
+
+When implementing a driver, the driver would be expected to export its UUID as part of the crate. It's likely the MnemOS project would maintain some registry of "known UUIDs", to attempt to avoid collisions. This could be in a text file, or wiki, etc.
+
+### Using `TypeId`
+
+Although all drivers are accessed by a `KChannel<T>`, which is a heap allocated structure, attempting to hold many channels of different `T`s in the same structure wouldn't work.
+
+Instead, the key:value map would need to type-erase the channel, likely just holding some kind of pointer. Theoretically, the UUID *should* be enough to ensure that the correct `T` matches the expected type, however the cost of getting this wrong (undefined behavior) due to a developer mistake is too high.
+
+In order to avoid this, the `get` function ALSO takes a `TypeId`, which is a unique type identifier provided by the compiler. Since all drivers are statically compiled together, they should all see the same `T` have the same `TypeId`.
+
+This WON'T necessarily work for userspace, or if MnemOS ever gets dynamically loaded drivers, which may be compiled with a different version of Rust, or potentially not Rust at all.
+
+For this reason, non-static users of these drivers will need to use the serialization interface instead. Userspace would send a serialized message to the expected UUID. The kernel would then use the associated `fn(&[u8]) -> Result<T>` to deserialize the message, and send it to the associated channel
+
+### Uniqueness of drivers
+
+This scheme requires that only one version of any kind of driver can be registered.
+
+This simplifies things, but could be potentially annoying for certain drivers are expected to have multiple instances, like keyboards, or storage devices.
+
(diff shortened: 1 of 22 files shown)


rfcs/0000-driver-registry.md
```diff
@@ -0,0 +1,379 @@
...
+There are also some half baked ideas in the `abi`/`mstd` crates about having a single "message kind" that is an enum of all possible driver types, but this isn't very extensible
+
+## The proposal
+
+This change proposes inntroducing a **registry** of drivers, maintained by the kernel.
+
+* This **registry** would be based on a Key:Value type arrangement:
+    * The **Key** would be a tuple of:
+        * A pre-defined **UUID**, that uniquely defines a "service" that the driver fulfills, e.g. the ability to process a given command type
+        * The `TypeId` of that message
```
hawkw, line 29: The intent behind this is to allow a single driver (with a single UUID) to handle multiple message types?

rfcs/0000-driver-registry.md
```diff
@@ -0,0 +1,379 @@
...
+
+There are a lot of different ways to discover drivers used in other operating systems, including using some kind of path, or a type erased file interface.
+
+This proposal takes an approach somewhat similar to Bluetooth's characteristics:
+
+* A UUID is chosen ahead of time to uniquely identify a service
+* Some "official" UUIDs are reserved for MnemOS and built-in drivers
+* A UUID represents a single service interface
+    * Although any driver could implement the UUID, all implementors are expected to have the same interface
+    * The UUID has no concept of versioning. If a breaking change is needed, a new UUID should be used
```
hawkw, line 86: my guess is that this would specifically be for breaking changes _in the driver's message-passing interface_ --- if the Rust API of a crate that provides a mnemOS driver changes, it wouldn't necessarily need to regenerate its UUID, unless the message-passing interface changes as well?

rfcs/0000-driver-registry.md
```diff
@@ -0,0 +1,379 @@
...
+
+For this reason, non-static users of these drivers will need to use the serialization interface instead. Userspace would send a serialized message to the expected UUID. The kernel would then use the associated deserialization function to deserialize the message, and send it to the associated channel
+
+### Uniqueness of drivers
+
+This scheme requires that only one version of any kind of driver can be registered.
+
+This simplifies things, but could be potentially annoying for certain drivers are expected to have multiple instances, like keyboards, or storage devices.
+
+This proposal sort of ignores that problem, and suggests that those drivers shouldn't be discoverable. Instead, some kind of "higher level" driver that hands out instances of those drivers should be registered.
```
hawkw, line 110: so, the expectation here would be something to the effect of...you have a "keyboards" driver that maintains a registry of any device that is a keyboard, and itself depends on individual drivers for specific hardware devices?

rfcs/0000-driver-registry.md
```diff
@@ -0,0 +1,379 @@
...
+
+Although there are other communication primitives available ready, such as the various flavors of the `bbq` based queues, they could be registered and provided through the KChannel as a "side channel". This is how the current (as of 2022-07-18) serial mux interface works:
+
+* The client who wants to open a virtual port gets the `KProducer`
+* The client sends a "register port" message
+* On success, the client gets back a `bbq` channel handle from the driver
+
+### Interfaces: Kernel Only
+
+If a type used in the channel does NOT support serialization/deserialization, it is limited to only being usable in the kernel, where this step is not required.
```
hawkw, line 124: it was not immediately clear from this summary that "kernel only" also includes other drivers in kernelspace, can we mention that here?

rfcs/0000-driver-registry.md
```diff
@@ -0,0 +1,379 @@
...
+## Interface: Userspace
+
+If a given type DOES implement Serialize/Deserialize, it can also be used for communication with userspace.
+
+The registry provides two additional functions in this case:
+
+```rust
+// This is a type that is capable of deserializing and processing
+// messages for a given UUID type
+impl UserspaceHandler {
```
hawkw, line 204: if a driver is provided by a separate crate, the `UserspaceHandler` type would just be some arbitrary type provided by that crate, right? or am i misinterpreting this?

hawkw: oh, nevermind, I see how it works --- the userspace handler is a type that knows how to do type-erased monomorphized deserialization.

i think it would be nice if this section summarized what a `UserspaceHandler` is up front, before going into detail on how it's implemented?

hawkw (commented):
this proposal makes sense overall and seems well thought out! i commented on some minor nits with the RFC --- there are some details that it might be good to expand on a bit more.

my only major design concern is that the registry does not handle situations where multiple instances of the same hardware device are present. instead, this is punted to the individual drivers to manage. i don't think this is necessarily an unacceptable design choice, but it does kind of feel like the main thing that we would want to think through all the consequences of.

some potential concerns include:
- if there end up being several drivers that can all manage multiple hardware device instances, each driver's message-passing API may end up implementing different schemes of routing requests to different devices.
  * a potential downside of this is that it creates a bunch of different interfaces that users have to program against
  * on the other hand, i suppose this _may_ be a good thing, as different hardware devices may have different requirements for how device instances are indexed?
 - potentially, there may be some efficiency concerns if a driver task is interested in _all_ instances of a hardware device, and wakes up when any of them has an I/O event? but, this could maybe be avoided by implementing main driver tasks as a sort of dispatcher that spawns worker tasks for each device instance and hands off requests from userspace to a bbqueue sidechannel directly to that device instance...

hawkw:
sad side note: looks like the [`uuid` crate](https://docs.rs/uuid/latest/uuid/index.html) does not support no-std. could be worth an upstream PR!

jamesmunns:
> sad side note: looks like the [`uuid` crate](https://docs.rs/uuid/latest/uuid/index.html) does not support no-std. could be worth an upstream PR!

That's a valid point, I wonder if it makes sense to just use a u64/u128 in place of a UUID and require an external crate. It would certainly be worth swapping over if 128 bits is necessary, not worried about breaking ABI changes yet.

hawkw:
> I wonder if it makes sense to just use a u64/u128 in place of a UUID and require an external crate. It would certainly be worth swapping over if 128 bits is necessary, not worried about breaking ABI changes yet.

I mean, a UUID is, essentially, [just a 128-bit value](https://docs.rs/uuid/latest/uuid/struct.Uuid.html#abi); the `uuid` crate mostly deals with things related to parsing and formatting UUIDs, as well as correctly constructing them.

It occurs to me that the `uuid` crate could still be used in _build scripts_ for driver crates to generate that driver's UUID (e.g. by SHA hashing the driver's name + version, to ensure that semver-incompatible releases generate a new UUID, or something?), and the resultant UUID could be exported by the driver crate as just a u128.

rfcs/0000-driver-registry.md
jamesmunns: I would say that if a crate, say `SerialPort` made breaking changes in it's declared (T, U) type, then a new UUID should be used. Otherwise we also need to capture a version number field for drivers as well.

rfcs/0000-driver-registry.md
jamesmunns: Yeah, I'd say you go one of two ways:

* A registered "SerialPortManager" driver, that hands out access to ONE of N serial ports on request through it's kproducer
* A registered "KeyboardManager" that takes clients, where all keyboards registered have their output mixed into a single broadcast stream.

jamesmunns: But really yeah, it isn't really prescribing one way or another, but I guess "consistent driver interface styles" can be done on a later stylistic pass?

rfcs/0000-driver-registry.md
```diff
@@ -0,0 +1,379 @@
...
+
+### (Exclusively) using `KChannel<T>`
+
+Although a lot of discussion of using "message passing" as a primary interface has been had so far, this would pretty much cement this design choice.
+
+Although there are other communication primitives available ready, such as the various flavors of the `bbq` based queues, they could be registered and provided through the KChannel as a "side channel". This is how the current (as of 2022-07-18) serial mux interface works:
+
+* The client who wants to open a virtual port gets the `KProducer`
+* The client sends a "register port" message
+* On success, the client gets back a `bbq` channel handle from the driver
```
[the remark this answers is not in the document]
jamesmunns: I see that as sort of outside the scope of this RFC! Proposing a new abi-compatible interface would be in another RFC, technically the only fully established interface is the "main ring".

I do want to do that later, do you think it makes sense weighing in on "future ideas" for after this RFC?

rfcs/0000-driver-registry.md
jamesmunns: Do you think a docs update is enough? e.g. in the rfc and method docs for `set_konly`, or does that need another name as well?

rfcs/0000-driver-registry.md
jamesmunns: Yeah, I give some hints into this in the code snippet ("show by usage"), but I should probably have some prose to go with it as well.

jamesmunns:
Misc thought (might add it to the RFC?), but I think this is how most drivers/crates would actually expose the functionality

```rust
//! A basic loop back on port 100

// Holds the types and uuid constant
use drivers::SerialPorts;

// Get the serial port mux interface
let ports = SerialPorts::new(&registry).await?;

// Open up port 0, gives us back a bidi-bbq
let port_0 = ports.open(100).await?;

// Spawn a "loopback port" task
kernel.spawn(async move {
    loop {
        let mut rgr = port_0.reader().read_grant().await;
        let len = rgr.len();

        port_0.send(&rgr).await;
        rgr.release(len);
    }
}).await;
```

jamesmunns:
Feedback seems pretty overall positive so far (thank you @hawkw!), I'm going to go ahead and start implementing this, and will update the RFC with anything I discover on the way.

It's a little out of order process wise, but I'll definitely come back and discuss if the impl doesn't match the original intent.

hawkw (approved):
fwiw, i'm fine with merging the RFC PR now so that the current design is documented, and updating it if you make changes while working on the implementation. but, keeping it in a branch until the impl merges is also fine, i just didn't want to hold it up.

PR #28, opened 2022-07-22
Implementation of Driver Registry RFC

jamesmunns:
Provides an implementation of #25.

There aren't any major conceptual changes here, but some implementation tweaks were made.

Additionally, I've gone ahead and "fully broken" userspace comms (the existing structures don't totally fit, and fixing them would require fixing some other underlying details, which I will do as a follow-up in a later RFC to restore userspace comms). The largest pre-req for this fix is deciding how to do async/await notifications across the ABI barrier.


jamesmunns:
As a note, this is what the current demo main now looks like, which I think is much closer to what I want to be possible:

https://github.com/tosc-rs/mnemos/blob/307b2e1283b724a3e3981cf8164b319c89429812/source/melpomene/src/main.rs#L82-L130

source/kernel/src/registry/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+pub static KNOWN_UUIDS: &[Uuid] = &[
+    //
+    // Kernel UUIDs
+    //
+
+    // SerialMux
+    uuid!("54c983fa-736f-4223-b90d-c4360a308647"),
+    // Simple Serial Port
+    uuid!("f06aac01-2773-4266-8681-583ffe756554"),
+];
```
hawkw, line 24: is an array of UUIDs really that useful? it seems like we might want to define this as a module with a bunch of constants in it, as well as an array. maybe something like this:
```suggestion
/// A partial list of known UUIDs of driver services
pub mod known_uuids {
     /// Kernel UUIDs
     pub mod kernel {
         pub const SERIAL_MUX: Uuid = uuid!("54c983fa-736f-4223-b90d-c4360a308647");
         pub const SIMPLE_SERIAL_PORT: Uuid = uuid!("f06aac01-2773-4266-8681-583ffe756554");
     }
     
     // In case you need to iterate over every UUID
     pub static ALL: &[Uuid] = &[
         kernel::SERIAL_MUX,
         kernel::SIMPLE_SERIAL_PORT,
     ];
}
```

source/kernel/src/registry/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+
+/// The driver registry used by the kernel.
+pub struct Registry {
+    items: HeapFixedVec<RegistryItem>,
+}
+
+// TODO: This probably goes into the ABI crate, here is fine for now
+#[derive(Serialize, Deserialize)]
+pub struct UserRequest<'a> {
+    // TODO: Maybe not the UUID, maybe pre-discover a shorter UID?
```
hawkw, line 56: we could maybe avoid the linear search of the registry if we had a way for userspace to look up the array index of a driver by its UUID, or something? then, you could do the linear search once for driver discovery, and then reference the driver by an index every time you want to talk to it.

source/kernel/src/registry/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+
+    /// This is the type of the response sent FROM the driver service
+    type Response: 'static;
+
+    /// This is the UUID of the driver service
+    const UUID: Uuid;
+
+    /// Get the type_id used to make sure that driver instances are correctly typed.
+    /// Corresponds to the same type ID as `(Self::Request, Self::Response)`
+    fn type_id() -> TypeId {
```
hawkw, line 43: we might want to make this function return the `TypeId` wrapped in a type without a public constructor, or something, in order to prevent user code from overriding this method?

source/kernel/src/registry/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+/// A `ReplyTo` is used to allow the CLIENT of a service to choose the
+/// way that the driver SERVICE replies to us. Essentially, this acts
+/// as a "self addressed stamped envelope" for the SERVICE to use to
+/// reply to the CLIENT.
+pub enum ReplyTo<U> {
+    // This can be used to reply directly to another kernel entity,
+    // without a serialization step
+    KChannel(KProducer<Envelope<Result<U, ()>>>),
+
+    Rosc(Sender<Envelope<Result<U, ()>>>),
```
hawkw, line 100: is it possible to send a `Rosc` reply to userspace, or is that kernel only as well?

source/kernel/src/comms/rosc.rs
```diff
@@ -0,0 +1,185 @@
...
+use crate::Kernel;
+
+/// A reusable One-Shot channel.
+///
+/// A `Rosc<T>` can be used to hand out single-use [Sender] items, which can
+/// be used to make a single reply.
+///
+/// A given `Rosc<T>` can only ever have zero or one `Sender<T>`s live at any
+/// given time, and a response can be received through a call to [Rosc::receive].
+pub struct Rosc<T> {
```
hawkw, line 29: style nit, take it or leave it: i personally don't love that this is an acronym; if i see `Rosc` for the first time in other code, i'm going to have to go look at the declaration before i know what the hell a `Rosc` is...

source/kernel/src/drivers/serial_mux.rs
```diff
@@ -28,40 +41,154 @@ pub enum Response {
...
+    incoming: bbq::Consumer,
+    mux: HeapArc<Mutex<MuxingInfo>>,
+}
+
+// impl SerialMux
+
+impl RegisteredDriver for SerialMux {
+    type Request = Request;
+    type Response = Response;
+    const UUID: Uuid = uuid!("54c983fa-736f-4223-b90d-c4360a308647");
```
hawkw, line 73: nit: it would probably be good if the UUID in the `registry` module's list of well-known UUIDs referenced this constant, instead of the same string occurring in two places...

source/kernel/src/registry/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+pub struct UserResponse<U> {
+    // TODO: Maybe not the UUID, maybe pre-discover a shorter UID?
+    uuid: Uuid,
+    nonce: u32,
+    reply: Result<U, ()>,
+}
+
+/// A wrapper for a message TO and FROM a driver service.
+/// Used to be able to add additional message metadata without
+/// changing the fundamental message type.
```
hawkw, line 74: does that mean that the intention is to add more fields to the `Envelope` type? if so, it should probably be
```suggestion
/// Used to be able to add additional message metadata without
/// changing the fundamental message type.
#[non_exhaustive]
```

source/kernel/src/registry/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+    ) -> Result<(), ()> {
+        if self.items.iter().any(|i| i.key == RD::UUID) {
+            return Err(());
+        }
+        self.items
+            .push(RegistryItem {
+                key: RD::UUID,
+                value: RegistryValue {
+                    req_resp_tuple_id: RD::type_id(),
+                    req_prod_leaked: kch.clone().leak_erased(),
```
hawkw, line 174: `leak_erased` sounds scary! why do we need to leak these, and do we need to ensure that these are eventually cleared up?

source/kernel/src/registry/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+            })
+            .map_err(drop)
+    }
+
+    /// Register a driver service for use in the kernel (including drivers) as
+    /// well as in userspace.
+    ///
+    /// See [Registry::set_konly] if the request and response types are not
+    /// serializable.
+    pub fn set<RD>(&mut self, kch: &KProducer<Message<RD>>) -> Result<(), ()>
```
hawkw, line 186: naming nit, take it or leave it: the name `set` (and `set_konly`) kind of imply to me that it can overwrite a previous value, but this fails if a driver is already registered. maybe consider `register` and `register_konly`, instead, so that it doesn't sound like a map where a value can be set multiple times?

hawkw, line 186: tiny style nit: might be nice if this returned an error type other than `()`; could be an empty marker type or at least a `&'static str` saying "cannot register a driver multiple times" or something? that way, `unwrap()` calls will have messages that are meaningful...

source/kernel/src/registry/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+
+// ReplyTo
+
+impl<U> ReplyTo<U> {
+    pub async fn reply_konly(self, payload: Result<U, ()>) -> Result<(), ()> {
+        let hmsg = Envelope { body: payload };
+        match self {
+            ReplyTo::KChannel(kprod) => kprod.enqueue_async(hmsg).await.map_err(drop),
+            ReplyTo::Rosc(sender) => sender.send(hmsg),
+            ReplyTo::Userspace { .. } => Err(()),
```
hawkw, line 273: nit, would be nice if error values were not discarded here; the error from a failed enqueue on a `KChannel` is _very_ different from "you tried to `reply_konly` to a `Userspace` message", and i would expect drivers to be able to want to handle those separately?

source/kernel/src/registry/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+        }
+    }
+}
+
+impl<U> ReplyTo<U>
+where
+    U: Serialize + MaxSize,
+{
+    pub async fn reply(self, uuid_source: Uuid, payload: Result<U, ()>) -> Result<(), ()> {
```
hawkw, line 282: again it kind of feels like a lot of error information is discarded here that maybe shouldn't be?

source/kernel/src/registry/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+/// TODO: I don't really know what to do with this. This is essentially
+/// declaring the client interface of a "Simple Serial" driver, without
+/// providing a definition of the actual driver service. In the simulator,
+/// this interface is used for the TcpSerial driver. This sort of "forward
+/// declaration" is needed when a driver in the kernel (like the SerialMux)
+/// depends on some external definition.
+///
+/// For non-kernel-depended services, it should be enough to depend on the
+/// actual driver you are consuming
+pub mod simple_serial {
```
hawkw, line 389: nit, tioli: should this maybe go in a `kernel::drivers` module or something, instead of in `registry`?

source/kernel/src/registry/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+            .await
+            .map_err(drop)
+    }
+}
+
+// -- other --
+
+/// A monomorphizable function that allows us to store the serialization type within
+/// the function itself, allowing for a type-erased function pointer to be stored
+/// inside of the registry.
```
hawkw, line 343: nit: would be nice if there was a mention of this fn's safety invariants in this comment?

hawkw (commented):
overall, this looks solid! i commented on some very minor nits; i'd like to look a little more closely at some of the implementation, but it looks good to me!

source/kernel/src/comms/rosc.rs
```diff
@@ -0,0 +1,185 @@
...
+    inner: HeapArc<Inner<T>>,
+}
```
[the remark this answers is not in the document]
jamesmunns: Yeah, I was thinking about this. I don't have a great answer, but would be open to porting it if it makes sense in the future.

source/kernel/src/lib.rs
```diff
@@ -197,41 +139,32 @@ impl Kernel {
...
+        let _k2u: FrameProducer<'static> = unsafe { BBBuffer::take_framed_producer(k2u_buf) };
+
+        #[allow(unreachable_code)]
+        if let Some(mut _reg) = self.registry.try_lock() {
+            // Incoming messages
+            while let Some(msg) = u2k.read() {
+                match postcard::from_bytes::<UserRequest>(&msg) {
+                    Ok(_req) => {
+                        // let kind = req.driver_kind();
+                        // if let Some(drv) = inner_mut.drivers.iter().find(|drv| drv.kind == kind) {
```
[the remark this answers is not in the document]
jamesmunns: *for now*, the number of drivers is likely to be small, and the "search" mechanism is an implementation detail.

I'd be open to a tree, or keeping the list sorted on insertion and using a binary search or something (since it still lives in linear memory).

source/kernel/src/lib.rs
```diff
@@ -240,6 +173,16 @@ impl Kernel {
         // TODO: Send time to userspace?
     }
 
+    // TODO: This prooooobably should instead use a joinhandle, and poll on the initialize future
+    // to completion, to make sure that certain actions actually complete.
```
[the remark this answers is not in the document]
jamesmunns: :heart:

source/kernel/src/registry/mod.rs
jamesmunns: Opened https://github.com/tosc-rs/mnemos/issues/32 to track this, I'd like to keep it simple for now and try using the registry before we spend time optimizing it (you make a very good point though!)

source/kernel/src/registry/mod.rs
jamesmunns: Rosc is kernel only, I'll update the comment.

source/kernel/src/registry/mod.rs
jamesmunns: These are not really "leaked" anymore with the vtable. so I've changed the name, thanks!

PR #34, opened 2022-07-25
feat(kernel): Add message IDs to make it easier to trace messages

jamesmunns:
This adds a number of `u32` ID fields to each message in order to make tracing interactions between services easier.

`u32` might be too much (I suspect u16 is probably enough, at least for service/client IDs), but I think this is a fine place to start. Most of the metadata insertion is "hidden" from end-consumers, happening in the built-in client and service handles, which means it should be hard to get wrong or forget to do.

### DIFF
--- a/source/kernel/src/registry/mod.rs
+++ b/source/kernel/src/registry/mod.rs
@@ -4,7 +4,7 @@ use mnemos_alloc::{containers::HeapFixedVec, heap::HeapGuard};
 use postcard::experimental::max_size::MaxSize;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use spitebuf::EnqueueError;
-use tracing::{info, debug};
+use tracing::{debug, info};
 use uuid::{uuid, Uuid};
 
 use crate::comms::{
@@ -269,7 +269,8 @@ impl Registry {
         if self.items.iter().any(|i| i.key == RD::UUID) {
             return Err(RegistrationError::UuidAlreadyRegistered);
         }
-        let result = self.items
+        let result = self
+            .items
             .push(RegistryItem {
                 key: RD::UUID,
                 value: RegistryValue {
@@ -305,7 +306,8 @@ impl Registry {
         if self.items.iter().any(|i| i.key == RD::UUID) {
             return Err(RegistrationError::UuidAlreadyRegistered);
         }
-        let result = self.items
+        let result = self
+            .items
             .push(RegistryItem {
                 key: RD::UUID,
                 value: RegistryValue {
@@ -391,8 +393,17 @@ impl Registry {
 // Envelope
 
 impl<P> Envelope<P> {
-    pub fn new(body: P, service_id: u32, client_id: u32, request_id: u32) -> Self {
-        Envelope { body, service_id, client_id, request_id }
+    /// Create a response Envelope from a given request Envelope.
+    ///
+    /// Maintains the same Service ID and Client ID, and increments the
+    /// request ID by one.
+    pub fn reply_with<U>(&self, body: U) -> Envelope<U> {
+        Envelope {
+            body,
+            service_id: self.service_id,
+            client_id: self.client_id,
+            request_id: self.request_id + 1,
+        }
     }
 }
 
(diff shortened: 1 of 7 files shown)


source/kernel/src/registry/mod.rs
```diff
@@ -89,6 +107,9 @@ pub struct UserResponse<U, E> {
 #[non_exhaustive]
 pub struct Envelope<P> {
     pub body: P,
+    service_id: u32,
+    client_id: u32,
+    request_id: u32,
```
hawkw, line 112: seems like it might eventually be worth introducing newtypes for `ServiceId`, `ClientId`, and `RequestId`, to ensure these aren't confusable? but we can do that in a follow-up.

source/kernel/src/registry/mod.rs
```diff
@@ -81,6 +83,22 @@ pub struct UserResponse<U, E> {
...
+// UserResponse
+
+impl<U: MaxSize, E: MaxSize> MaxSize for UserResponse<U, E> {
+    //
+    // KEEP IN SYNC WITH STRUCT DEFINITION ABOVE!
+    //
+    const POSTCARD_MAX_SIZE: usize = {
+        <[u8; 16] as MaxSize>::POSTCARD_MAX_SIZE
+            + <u32 as MaxSize>::POSTCARD_MAX_SIZE
+            + <Result<U, E> as MaxSize>::POSTCARD_MAX_SIZE
```
hawkw, line 100: would be really cool if this was derivable (or even just declarative-macro-able)...i might try and hack that together for a subsequent branch...

source/kernel/src/registry/mod.rs
```diff
@@ -170,18 +191,25 @@ impl<T> From<EnqueueError<T>> for ReplyError {
...
+    client_id: u32,
+    request_id: u32,
 }
 
 type ErasedDeserHandler = unsafe fn(
     UserRequest<'_>,
     &ErasedKProducer,
     &bbq::MpscProducer,
+    u32,
+    u32,
```
hawkw, line 212: oh, yeah, we should _definitely_ newtype these...

source/kernel/src/registry/mod.rs
```diff
@@ -259,16 +306,21 @@ impl Registry {
...
                     req_resp_tuple_id: RD::type_id().type_of(),
                     req_prod: kch.clone().type_erase(),
                     req_deser: Some(map_deser::<RD>),
+                    service_id: self.counter,
                 },
             })
-            .map_err(|_| RegistrationError::RegistryFull)
+            .map_err(|_| RegistrationError::RegistryFull)?;
+        info!(uuid = ?RD::UUID, service_id = self.counter, "Registered");
+        self.counter = self.counter.wrapping_add(1);
```
hawkw, line 322: wonder if we should maybe log something in the event that this wraps, just so that a reader of the logs will know that they may see the same IDs repeat

source/kernel/src/registry/mod.rs
```diff
@@ -299,37 +363,47 @@ impl Registry {
...
-
 // Envelope
 
 impl<P> Envelope<P> {
-    pub fn new(body: P) -> Self {
-        Envelope { body }
+    /// Create a response Envelope from a given request Envelope.
+    ///
+    /// Maintains the same Service ID and Client ID, and increments the
+    /// request ID by one.
```
hawkw, line 399: i don't get this behavior...why does a response to a request increment the request ID? shouldn't the response habe the same req ID, so that we know what request it's a response to? or am i missing something?

hawkw (approved):
looks good, commented on a few minor suggestions

source/kernel/src/registry/mod.rs
jamesmunns: So, the macro is derivable, but the problem is that `UUID` is an extern type, so I'd have to do some newtype shenanigans on it to impl the MaxSize trait for it. I was being lazy, but that might be a better idea.

source/kernel/src/registry/mod.rs
jamesmunns: So, my GOAL was to have a predictable but different ID for replies, so they could be distinguished in logs as coming from a different sender. Right now I have client/service id, which doesn't flip on the "response". Essentially this was intended as a (hacky) way to use the lsb as a "request/response" flag.

It might have been poorly executed.

hawkw: Ah, okay, that makes sense, i think. But I would kind of expect any logs mentioning a response ID would also indicate that it's a response through some other mechanism?

source/kernel/src/registry/mod.rs
hawkw: ah, i didn't think of that part. fair point.

source/kernel/src/registry/mod.rs
jamesmunns: Kicking this can down the road.

PR #35, opened 2022-07-27
framebuf driver for embedded graphics

arunvijayshankar:
Added a framebuf driver for embedded graphics

### DIFF
--- a/source/kernel/Cargo.toml
+++ b/source/kernel/Cargo.toml
@@ -63,6 +63,3 @@ features = ["experimental-derive"]
 
 [dependencies.embedded-graphics]
 version = "0.7.1"
-
-[dependencies.embedded-graphics-simulator]
-version = "0.3.0"
--- a/source/kernel/Cargo.toml
+++ b/source/kernel/Cargo.toml
@@ -60,3 +60,9 @@ features = ["derive"]
 [dependencies.postcard]
 version = "1.0.1"
 features = ["experimental-derive"]
+
+[dependencies.embedded-graphics]
+version = "0.7.1"
+
+[dependencies.embedded-graphics-simulator]
+version = "0.3.0"
--- a/source/kernel/src/drivers/emb_display.rs
+++ b/source/kernel/src/drivers/emb_display.rs
@@ -15,9 +15,8 @@ use uuid::Uuid;
 use embedded_graphics::{
     pixelcolor::{Gray8, GrayColor},
     prelude::*,
-    image::{Image, ImageRaw},
+    image::ImageRaw,
 };
-use embedded_graphics_simulator::SimulatorDisplay;
 
 const BYTES_PER_PIXEL: u32 = 1;
 const DISP_WIDTH: u32 = 319;
@@ -134,13 +133,15 @@ impl EmbDisplay {
 // On a physical display, the raw pixel data can be sent over to the display directly 
 // Using the display's device interface
 impl FrameChunk {
-    pub async fn frame_display(&mut self) -> Result<SimulatorDisplay<Gray8>, ()> {
-        let mut sdisp = SimulatorDisplay::<Gray8>::new(Size::new(320, 240));
-        let raw_image = ImageRaw::<Gray8>::new(self.bytes.as_ref(), self.width);
-        let image = Image::new(&raw_image, Point::new(self.start_x, self.start_y));
-        image.draw(&mut sdisp).unwrap();
+    pub async fn frame_display(&mut self) -> Result<ImageRaw<Gray8>, ()> {
+        let raw_image: ImageRaw<Gray8>;
+        raw_image = ImageRaw::<Gray8>::new(self.bytes.as_ref(), self.width);
+        Ok(raw_image)
+    }
+
+    pub async fn frame_clear(&mut self) {
         for elem in self.bytes.iter_mut() { *elem = 0; }
-        Ok(sdisp)
+        
     }
 }
 
--- a/source/kernel/src/drivers/emb_display.rs
+++ b/source/kernel/src/drivers/emb_display.rs
@@ -57,11 +57,11 @@ pub struct EmbDisplayHandle {
 }
 
 pub enum Request {
-    NewFrameChunk { 
-        frame_id: u16, 
-        start_x: i32, 
-        start_y: i32, 
-        width: u32, 
+    NewFrameChunk {
+        frame_id: u16,
+        start_x: i32,
+        start_y: i32,
+        width: u32,
         height: u32,
     },
 }
@@ -76,7 +76,7 @@ pub enum EmbDisplayError {
 
 struct CommanderTask {
     cmd: KConsumer<Message<EmbDisplay>>,
-    fmutex: HeapArc<Mutex<DisplayInfo>>, 
+    fmutex: HeapArc<Mutex<DisplayInfo>>,
 }
 
 // impl EmbDisplay
@@ -105,7 +105,7 @@ impl EmbDisplay {
 
         let imutex = kernel
             .heap()
-            .allocate_arc(Mutex::new(DisplayInfo { 
+            .allocate_arc(Mutex::new(DisplayInfo {
                 kernel,
                 frames,
                 frame_size,
@@ -134,7 +134,7 @@ impl FrameChunk {
     ///
     /// Pixel data is turned into a raw image, and then drawn onto a SimulatorDisplay object
     /// This is necessary as a e-g Window only accepts SimulatorDisplay object
-    /// On a physical display, the raw pixel data can be sent over to the display directly 
+    /// On a physical display, the raw pixel data can be sent over to the display directly
     /// Using the display's device interface
     pub fn frame_display(&mut self) -> Result<ImageRaw<Gray8>, ()> {
         let raw_image: ImageRaw<Gray8>;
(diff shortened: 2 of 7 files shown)


source/melpomene/src/main.rs
```diff
@@ -135,6 +147,103 @@ fn kernel_entry(opts: MelpomeneOptions) {
...
+        let bline = Line::new(Point::new(148, 24), Point::new(8, 24))
+            .into_styled(line_style);
+        let lline = Line::new(Point::new(7, 24), Point::new(7, 2))
+            .into_styled(line_style);
+        let rline = Line::new(Point::new(149, 24), Point::new(149, 2))
+            .into_styled(line_style);
+        let text_style = MonoTextStyle::new(&FONT_6X9, Gray8::WHITE);
+        let datetime_style = MonoTextStyle::new(&FONT_5X7, Gray8::WHITE);
+        let text1 = Text::new("Welcome to mnemOS!", Point::new(10, 10), text_style);
+        let text2 = Text::new("A tiny operating System", Point::new(10, 20), text_style);
```
hawkw, line 173: typo:
```suggestion
        let text2 = Text::new("A tiny operating system", Point::new(10, 20), text_style);
```

arunvijayshankar: corrected. thanks!

jamesmunns:
Hey @arunvijayshankar, I took a look and this looks really good!

One question - is it possible to move the `simulator` components to ONLY exist in melpo, not the kernel?

The kernel needs to be `no_std`, which means it should only hold the `embedded-graphics` dependency.

source/kernel/Cargo.toml
```diff
@@ -60,3 +60,9 @@ features = ["derive"]
 [dependencies.postcard]
 version = "1.0.1"
 features = ["experimental-derive"]
+
+[dependencies.embedded-graphics]
+version = "0.7.1"
+
+[dependencies.embedded-graphics-simulator]
+version = "0.3.0"
```
hawkw, line 68: Looks like we forgot to remove the actual dependency on `embedded-graphics-simulator` from the kernel?

source/kernel/src/drivers/emb_display.rs
```diff
@@ -0,0 +1,260 @@
...
+    const UUID: Uuid = crate::registry::known_uuids::kernel::EMB_DISPLAY;
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum RegistrationError {
+    DisplayAlreadyExists,
+}
+
+// Register the driver instance
+impl EmbDisplay {
```
hawkw, line 93: nit: i think this comment was supposed to go on the method?
```suggestion
impl EmbDisplay {
    /// Register the driver instance
```

source/kernel/src/drivers/emb_display.rs
```diff
@@ -0,0 +1,260 @@
...
+    }
+}
+
+// Create and return a Simulator display object from raw pixel data.
+// Pixel data is turned into a raw image, and then drawn onto a SimulatorDisplay object
+// This is necessary as a e-g Window only accepts SimulatorDisplay object
+// On a physical display, the raw pixel data can be sent over to the display directly 
+// Using the display's device interface
+impl FrameChunk {
+    pub async fn frame_display(&mut self) -> Result<ImageRaw<Gray8>, ()> {
```
hawkw, line 136: nit: i think this comment is supposed to be on the method?
```suggestion
impl FrameChunk {
    /// Create and return a Simulator display object from raw pixel data.
    ///
    /// Pixel data is turned into a raw image, and then drawn onto a SimulatorDisplay object
    /// This is necessary as a e-g Window only accepts SimulatorDisplay object
    /// On a physical display, the raw pixel data can be sent over to the display directly 
    /// Using the display's device interface
    pub async fn frame_display(&mut self) -> Result<ImageRaw<Gray8>, ()> {
```

source/kernel/src/drivers/emb_display.rs
```diff
@@ -0,0 +1,260 @@
...
+        let raw_image: ImageRaw<Gray8>;
+        raw_image = ImageRaw::<Gray8>::new(self.bytes.as_ref(), self.width);
+        Ok(raw_image)
+    }
+
+    pub async fn frame_clear(&mut self) {
```
hawkw, line 142: as far as I can tell, neither of these methods _need_ to be async --- they don't `.await` anything. can we change them to be synchronous?
```suggestion
    pub fn frame_display(&mut self) -> Result<ImageRaw<Gray8>, ()> {
        let raw_image: ImageRaw<Gray8>;
        raw_image = ImageRaw::<Gray8>::new(self.bytes.as_ref(), self.width);
        Ok(raw_image)
    }

    pub fn frame_clear(&mut self) {
```

source/kernel/src/drivers/emb_display.rs
```diff
@@ -0,0 +1,260 @@
...
+            .await;
+        let (cmd_prod, cmd_cons) = KChannel::new_async(kernel, 1).await.split();
+        let commander = CommanderTask {
+            cmd: cmd_cons,
+            fmutex: imutex,
+        };
+
+        kernel.spawn(async move{
+            commander.run().await;
+        }).await;
```
hawkw, line 119: tiny style nit, take it or leave it: this creates an unnecessary `async` block. since all we're doing is calling an async fn and awaiting it, this could just be:
```suggestion
        kernel.spawn(commander.run()).await;
```

source/kernel/src/drivers/emb_display.rs
```diff
@@ -0,0 +1,260 @@
...
+    pixelcolor::{Gray8, GrayColor},
+    prelude::*,
+    image::ImageRaw,
+};
+
+const BYTES_PER_PIXEL: u32 = 1;
+const DISP_WIDTH: u32 = 319;
+const DISP_HEIGHT: u32 = 239;
+
+// Registered driver
```
hawkw, line 25: nit: is this supposed to be a doc comment?
```suggestion
/// Registered driver
```

source/kernel/src/drivers/emb_display.rs
```diff
@@ -0,0 +1,260 @@
...
+pub struct EmbDisplay {
+    kernel: &'static Kernel,
+}
+
+// FrameChunk is recieved after client has sent a request for one
```
hawkw, line 30: nit: is this supposed to be a doc comment?
```suggestion
/// FrameChunk is recieved after client has sent a request for one
```

source/kernel/src/drivers/emb_display.rs
```diff
@@ -0,0 +1,260 @@
...
+
+        let size = (width * height) as usize;
+        let bytes = self.kernel.heap().allocate_array_with(|| 0, size).await;
+        let fc = FrameChunk {
+            frame_id: frame_id,
+            bytes: bytes,
+            start_x: start_x,
+            start_y: start_y,
+            width: width,
+            height: height,
```
hawkw, line 204: style nit, not a big deal: we can use [field initializer shorthand](https://rust-lang.github.io/rfcs/1682-field-init-shorthand.html) here, which maybe looks a little nicer:
```suggestion
            frame_id
            bytes,
            start_x,
            start_y,
            width,
            height,
```

hawkw (commented):
this is really cool! i left a few minor suggestions, i hope they're helpful.

hawkw:
also, iit looks like [`rustfmt` needs to be run on this branch to make CI happy](https://github.com/tosc-rs/mnemos/runs/7587005883?check_suite_focus=true).

source/kernel/Cargo.toml
arunvijayshankar: oops! haha missed this one!

source/kernel/src/drivers/emb_display.rs
arunvijayshankar: makes sense!

source/kernel/src/drivers/emb_display.rs
arunvijayshankar: yes, that makes sense. Sorry, this should have been obvious to me :)

source/kernel/src/drivers/emb_display.rs
arunvijayshankar: ah ok! what would be the use case for retaining the outer async block? Sorry for the basic question. I'm very new to rust and asynchronous programming! Thanks for the suggestions :)

source/kernel/src/drivers/emb_display.rs
arunvijayshankar: This is just a comment I think. I retained the same style of comments from serial_mux.rs

source/kernel/src/drivers/emb_display.rs
arunvijayshankar: Also just a standard comment. I should probably put in document comments, come to think of it

source/kernel/src/drivers/emb_display.rs
arunvijayshankar: This does look  a lot better!

source/kernel/src/drivers/emb_display.rs
hawkw: i don't really think there's a reason you would want to do it — since there's no other code in the async block, you can just spawn the method call.

hawkw (approved):

jamesmunns:
As a note @arunvijayshankar, you can run `cargo fmt --all` from the root of the project, and it'll format all of the crates :)

PR #36, opened 2022-07-28
feat(platform): Allwinner D1 Support

jamesmunns:


source/kernel/src/lib.rs
```diff
@@ -202,7 +202,8 @@ impl Kernel {
     }
 
     pub fn spawn_allocated<F: Future + 'static>(&'static self, task: HeapBox<Task<F>>) {
-        self.inner.scheduler.spawn_allocated::<F, HBStorage>(task)
+        // TODO: Join Handle support!
+        let _ = self.inner.scheduler.spawn_allocated::<F, HBStorage>(task);
```
hawkw, line 206: i believe the TODO is as simple as changing this to:
```rust
    pub fn spawn_allocated<F: Future + 'static>(&'static self, task: HeapBox<Task<F>>) -> JoinHandle<F> {
        self.inner.scheduler.spawn_allocated::<F, HBStorage>(task)
    }
```
(and an appropriate import somewhere)

source/platforms/allwinner-d1/src/main.rs
```diff
@@ -0,0 +1,623 @@
...
+macro_rules! println {
+    ($($arg:tt)*) => {
+        $crate::_print(core::format_args!($($arg)*));
+        $crate::print!("\r\n");
+    }
+}
+
+static TICK_MS: AtomicU64 = AtomicU64::new(0);
+static TICK_WAKER: AtomicPtr<WaitCell> = AtomicPtr::new(null_mut());
+static UART_TX_WAKER: AtomicPtr<WaitCell> = AtomicPtr::new(null_mut());
```
hawkw, line 71: FWIW, `WaitCell::new` is a const fn, so I _think_ that rather than allocating `HeapArc`s and storing them in an `AtomicPtr`, you could just have a
```suggestion
static TICK_WAKER: WaitCell = WaitCell::new();
static UART_TX_WAKER: WaitCell: WaitCell::new();
```

(of course, this would mean modifying other code as well)

source/platforms/allwinner-d1/src/main.rs
```diff
@@ -0,0 +1,623 @@
...
+        }
+    }
+}
+
+impl TimerQueue {
+    pub async fn register(kernel: &'static Kernel) -> Result<HeapArc<WaitCell>, ()> {
+        let wait = kernel.heap().allocate_arc(WaitCell::new()).await;
+        let (kprod, kcons) = KChannel::new_async(kernel, 32).await.split();
+        let arr = kernel.heap().allocate_array_with(|| None, 128).await;
+        let arr = kernel.heap().allocate_arc(Mutex::new(arr)).await;
```
hawkw, line 551: since this is a fixed size array, and `maitake::sync::Mutex::new` is a const fn, couldn't this just be a static rather than an arc?
```rust
static TIMER_QUEUE_ARRAY: Mutex<[Option<(u64, Message<TimerQueue>)>, 128]> = Mutex::new([None; 128]);
```
should work...

source/platforms/allwinner-d1/src/plic.rs
```diff
@@ -0,0 +1,139 @@
...
+            22 => Ok(Priority::P22),
+            23 => Ok(Priority::P23),
+            24 => Ok(Priority::P24),
+            25 => Ok(Priority::P25),
+            26 => Ok(Priority::P26),
+            27 => Ok(Priority::P27),
+            28 => Ok(Priority::P28),
+            29 => Ok(Priority::P29),
+            30 => Ok(Priority::P30),
+            31 => Ok(Priority::P31),
```
hawkw, line 135: this has big "maybe this should just be a newtype around a u32" energy, imo...

hawkw (commented):
this is extremely cool! some drive by suggestions

source/kernel/src/lib.rs
jamesmunns: Yeah. that makes sense! I was just fast-forwarding the dep to get the waitcell fix :D

source/platforms/allwinner-d1/src/main.rs
jamesmunns: Oh, duh, good call!

source/platforms/allwinner-d1/src/main.rs
jamesmunns: Also yes!

jamesmunns:
Superseded by #69.

PR #41, opened 2023-05-22
RFC: a forth userspace

jamesmunns:
[Rendered](https://github.com/tosc-rs/mnemos/blob/2023-a-new-beginning/rfcs/0000-a-forth-userspace.md)

### DIFF
--- a/rfcs/0000-a-forth-userspace.md
+++ b/rfcs/0000-a-forth-userspace.md
@@ -2,7 +2,7 @@
 
 ## Goal
 
-In order to make early prototyping and iterating on the design of mnemos, this RFC proposes the
+In order to make early prototyping and iterating on the design of mnemos **easier**, this RFC proposes the
 addition of a single kind of "userspace" environment, a `forth` vm that can act as a program
 or interactive shell interface. This RFC also proposes a minimal set of capabilities and interfaces
 necessary to make this possible.
--- a/rfcs/0000-a-forth-userspace.md
+++ b/rfcs/0000-a-forth-userspace.md
@@ -52,8 +52,7 @@ this environment is likely suitable for two main things:
 * We'll need to add some kind of pattern/interface for managing cooperative userspace tasks
     * We will probably NOT enforce preemptive multitasking for now
     * We will probably NOT enforce memory protection/mapping for now
-    * userspace tasks yield when they feel like it, kernel then serves each tasks's executor like a metaexecutor
-    * userspace tasks stay sleeping as long as nothing in the kernel sends them messages
+    * userspace tasks yield when they feel like it, maybe with some "VM fuel" cap to force cooperation
 * We'll need some sort of "trampoline" code to bolt the forth3 vm to whatever userspace tasks look like
     * Start the vm, get it running
     * Alloc dictionary fragments as necessary
--- a/rfcs/0000-a-forth-userspace.md
+++ b/rfcs/0000-a-forth-userspace.md
@@ -25,11 +25,11 @@ this environment is likely suitable for two main things:
 
 * The OS will spawn a single userspace entity at boot time
     * This entity will be a cooperative async task running an interactive forth vm
-    * It will be capable of interacting the userspace/kernel interface (sending/receiving kernel messages)
+    * It will be capable of interacting with the userspace/kernel interface (sending/receiving kernel messages)
 * This "task 0" forth environment will be capable of spawning other userspace tasks
     * It can only spawn other forth vms, running their own separate environments
-    * A snapshot of its dictionary will be made and shared with the child environment
-    * All forth vm tasks will be cooperative
+    * When spawning a new task, a snapshot of the current task's dictionary will be made and shared with the child environment
+    * All forth vm tasks will be cooperatively scheduled
 
 ## Details
 
--- a/rfcs/0000-a-forth-userspace.md
+++ b/rfcs/0000-a-forth-userspace.md
@@ -41,6 +41,12 @@ this environment is likely suitable for two main things:
     * todo: is this all we need it for? what about variable addrs in variables?
 * We'll need some way of allocating + deepcopying dictionary fragments
 
+**Note**, some of these changes are already in-progress at the time of writing of this RFC:
+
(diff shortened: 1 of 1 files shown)


rfcs/0000-a-forth-userspace.md
```diff
@@ -0,0 +1,72 @@
...
+    * good practice for later, more general, "spawn" interface
+* Call the "spawn" service with our `task-id-0`, giving it some reasonable stdin/stdout
+
+## Other motivations and thoughts
+
+* Having something useful to play with NOW is more important than designing something perfect
+* We will learn a lot about what "spawn" needs to do with a relatively restricted interface
+    * We own the VM, we can force it to cooperate before we have preemption
+    * We can impl memory safety checks (probably not fortified, but good for "oops" checking) before we have MMU support
+    * All code is interpreted with the same vm engine, no need to worry about relocations, elf loading, or memory mapping yet
```
hawkw, line 72: technically, my current plan for spawning will create new instances of the `Forth` type from `forth3`, so if that's what you mean by "same VM engine", this isn't technically correct. but if you mean just by the same _kind_ of VM, yes!

rfcs/0000-a-forth-userspace.md
```diff
@@ -0,0 +1,72 @@
...
+    * like obtaining the address of a variable
+    * todo: is this all we need it for? what about variable addrs in variables?
+* We'll need some way of allocating + deepcopying dictionary fragments
+
+### Changes to `mnemos`
+
+* We'll need to add some kind of pattern/interface for managing cooperative userspace tasks
+    * We will probably NOT enforce preemptive multitasking for now
+    * We will probably NOT enforce memory protection/mapping for now
+    * userspace tasks yield when they feel like it, kernel then serves each tasks's executor like a metaexecutor
```
hawkw, line 49: not totally sure what you mean by "serves each task's executor like a metaexacutor", exactly --- each forth VM would just get a single task in the kernel's maitake scheduler, for now

rfcs/0000-a-forth-userspace.md
```diff
@@ -0,0 +1,72 @@
...
+* We'll need to add some kind of allocation service for bigpages
+    * Define a service
+    * Needs a way for allocating unshared r/w pages
+    * Needs a way to exchange unshared r/w pages for shared r/o pages
+    * These would *normally* be used to add memory slabs to the userspace heap allocator
+* We need some sort of "spawn" service
+    * ONLY for spawning forth vms now
+    * good practice for later, more general, "spawn" interface
+* Call the "spawn" service with our `task-id-0`, giving it some reasonable stdin/stdout
+
```
hawkw, line 65: next steps: IPC between forth tasks (maybe just start with waiting for a wakeup/waiting for a task to complete, and then sending data once that works). maybe that deserves a new RFC when we get there.

rfcs/0000-a-forth-userspace.md
```diff
@@ -0,0 +1,72 @@
...
+    * This entity will be a cooperative async task running an interactive forth vm
+    * It will be capable of interacting with the userspace/kernel interface (sending/receiving kernel messages)
+* This "task 0" forth environment will be capable of spawning other userspace tasks
+    * It can only spawn other forth vms, running their own separate environments
+    * When spawning a new task, a snapshot of the current task's dictionary will be made and shared with the child environment
+    * All forth vm tasks will be cooperatively scheduled
+
+## Details
+
+### Changes to `forth3` - the current forth vm
```
hawkw, line 36: maybe worth linking my in progress branch on forth3 for this?

hawkw (approved):
this is great, it definitely captures most of the salient points from our conversation earlier --- thanks for taking the time to write it down! left a couple nitpicks but overall i think this is fantastic!

rfcs/0000-a-forth-userspace.md
jamesmunns: Yeah! The code living in `.text` is the same for every app - there's no runtime loading or remapping. But yes, there are multiple instances of `self` (but not `Self::process_line` if that makes sense).

rfcs/0000-a-forth-userspace.md
jamesmunns: Yeah, I've modeled kernel-service:kernel-service comms, and kernel-service:userspace comms, but haven't designed any userspace:userspace comms yet.

Erlang has some interesting thoughts here (using unique task ids, and using those as routable source/destination addresses for messages).

rfcs/0000-a-forth-userspace.md
jamesmunns: Yeah, I guess I made two steps here:

At first, the kernel will just run tasks "raw" in kernel space.

Later, when there is real control separation between kernel and userspace, I expect each userspace "process" to have it's own executor. I sorta plan to expose userspace processes as futures in the kernel - but polling them yields control flow to the userspace program - essentially handing control to the subordinate executor, until either it yields (no pending tasks), the outgoing ringbuffer is full, etc.

A fair amount of this is probably worth a whole separate RFC later, but this is where my head was at when writing that.

rfcs/0000-a-forth-userspace.md
hawkw: right, yeah.

rfcs/0000-a-forth-userspace.md
jamesmunns: Linked!

rfcs/0000-a-forth-userspace.md
hawkw: i think that makes sense, when userspace programs can be compiled Rust realcodes as well as interpreted forth programs! i was just thinking in the context of the forth stuff. would love to see your thoughts on the ABI for gluing an async rust binary into the OS' scheduler, eventually!

rfcs/0000-a-forth-userspace.md
hawkw: that probably deserves its own RFC once we get there!

rfcs/0000-a-forth-userspace.md
jamesmunns: Yeah, I agree that's better to save for another day. I have lots of ideas, some of them might be good. Not strictly necessary now :)

PR #42, opened 2023-05-27
Integrate console loopback demo

jamesmunns:


jamesmunns:
Can I embed youtube stuff?

<iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/p23mpdcAZOY" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>

source/melpomene/src/main.rs
```diff
@@ -153,77 +152,110 @@ fn kernel_entry(opts: MelpomeneOptions) {
...
-        let output_settings = OutputSettingsBuilder::new()
-            .theme(BinaryColorTheme::OledBlue)
-            .build();
-
-        let mut sdisp = SimulatorDisplay::<Gray8>::new(Size::new(320, 240));
-        let mut window = Window::new("mnemOS", &output_settings);
+        let char_y = PROFONT_12_POINT.character_size.height;
+        let char_x = PROFONT_12_POINT.character_size.width + PROFONT_12_POINT.character_spacing;
+
+        // Draw titlebar
```
hawkw, line 170: not a blocker, just a thought for the future: it seems like a bunch of this stuff ought to move out of melpomene and into the platform agnostic kernel crate? it would be nice if the kernel could get a display driver handle from the driver registry and draw the shell and stuff, instead of doing it in melpomene. 

i guess we would probably want to be doing stuff like determining what font size to use based on the display resolution instead of hard coding it, so it might be a bit more complicated. none of this is a blocker for merging this change, just a note for the future.

source/melpomene/src/main.rs
```diff
@@ -153,77 +152,110 @@ fn kernel_entry(opts: MelpomeneOptions) {
...
-                        sdisp.clear(Gray8::BLACK).unwrap();
+                loop {
+                    // Wait until there is a frame buffer ready. There wouldn't be if we've spammed frames
+                    // before they've been consumed.
+                    let mut fc_0 = loop {
+                        let fc = disp_hdl
+                            .get_framechunk(0, char_y as i32, 400, 240 - char_y)
+                            .await;
+                        if let Some(fc) = fc {
+                            break fc;
```
hawkw, line 229: i was wondering if this could be expressed nicer as a `while` loop, but i guess it would require a `mut Option<FrameChunk>` outside of the loop, so 🤷‍♀️

source/melpomene/src/main.rs
```diff
@@ -153,77 +152,110 @@ fn kernel_entry(opts: MelpomeneOptions) {
...
+                    let rgr = p2.consumer().read_grant().await;
+                    for b in rgr.iter() {
+                        match rline.append_local_char(*b) {
+                            Ok(_) => {}
+                            // backspace
+                            Err(_) if *b == 0x7F => {
+                                rline.pop_local_char();
+                            }
+                            Err(_) if *b == b'\n' => {
+                                rline.submit_local_editing();
```
hawkw, line 246: it feels a little bit weird to me that we always try to append a character, and then if we get an error, we handle the backspace and newline cases. i feel like it might be a bit easier to follow if we matched on `b`, called `pop_local_char` if it's backspace, called `submit_local_char` if it's `\n`, and called `append_local_char` otherwise? 

this isn't a huge deal to me though, just a thought.

source/melpomene/src/main.rs
```diff
@@ -153,77 +152,110 @@ fn kernel_entry(opts: MelpomeneOptions) {
...
+                Point::new(0, PROFONT_12_POINT.baseline as i32),
+                text_style,
+            );
+            text1.draw(&mut fc_0).unwrap();
+
+            let title = "forth shell";
+            let text2 = Text::new(
+                title,
+                Point::new(
+                    400 - ((title.len() as u32) * char_x) as i32,
```
hawkw, line 185: maybe worth making the `400` that's referenced in a few places a constant so we don't have to update it all over the place if we change the size of the display?

hawkw (approved):
looks great overall! i had a few minor style suggestions

source/melpomene/src/main.rs
jamesmunns: Yeah, this is probably worth a think about how to split this up. Right now the only impl is in melpomene, and there's no real "standard init process": the kernel is more like a grab bag of lib components that you need to glue together.

I think we should look at this again when we start moving bits from the simulator into the kernel itself so we can reuse them on the D1 hardware.

source/melpomene/src/sim_drivers/emb_display.rs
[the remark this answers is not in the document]
jamesmunns: There were some pretty significant changes. The old version didn't do the e-g-sim rendering in the driver itself, and so the "application" sorta directly drove everything. This moved a large part of the logic into the driver server.

source/melpomene/src/main.rs
```diff
@@ -153,77 +152,110 @@ fn kernel_entry(opts: MelpomeneOptions) {
...
+                                rline.pop_local_char();
+                            }
+                            Err(_) if *b == b'\n' => {
+                                rline.submit_local_editing();
+                                b"And the computer says hi.".iter().for_each(|b| {
+                                    let _ = rline.append_remote_char(*b);
+                                });
+                                rline.submit_remote_editing();
+                            }
+                            Err(_) => {
```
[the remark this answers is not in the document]
jamesmunns: Err would fire on the line being too full, or an ascii control character being sent. I'll add a log for it.

source/melpomene/src/main.rs
jamesmunns: Yeah, I guess I know that it is already checking for "good ascii character" inside of "append", so it seemed a little silly to check before then check again inside the library.

We might want more robust input handling in the future (like handling more control characters properly), I might punt on that for now.

source/melpomene/src/main.rs
jamesmunns: That's a good call. Let me add some consts for that.

PR #48, opened 2023-06-01
WIP: "bag of holding" abstraction pattern

jamesmunns:


jamesmunns:
@hawkw this is ready for review IMO, but relies on https://github.com/jamesmunns/forth3/pull/19 being landed first

source/kernel/src/forth.rs
```diff
@@ -191,6 +299,147 @@ impl Params {
...
+    ///
+    /// This will only succeed if the same `T` is used as was used when calling
+    /// [`BagOfHolding::register()`], and if the token matches the one returned
+    /// by `register()`.
+    ///
+    /// If the token is unknown, or the `T` does not match, `None` will be returned
+    ///
+    /// At the moment, no `get_mut` functionality is provided. It *could* be, as the
+    /// Bag of Holding represents ownership of the contained items, however it would
+    /// not be possible to retrieve multiple mutable items at the same time. This
```
hawkw, line 404: is retrieving multiple bagged items mutably at the same time something we actually want to do? AFAICT, each bag of holding is uniquely owned by a forth VM instance, right?

source/kernel/src/forth.rs
```diff
@@ -191,6 +299,147 @@ impl Params {
...
+        unsafe { Some(t.as_ref()) }
+    }
+}
+
+/// A container item for a type-erased object
+struct BohValue {
+    /// The type id of the `T` pointed to by `leaked`
+    tid: TypeId,
+    /// A non-null pointer to a `T`, contained in a leaked `HeapBox<T>`.
+    leaked: NonNull<()>,
```
hawkw, line 425: nit, take it or leave it: i don't love the name `leaked` for this, since it's...not actually leaked, really. i might just call it `value` or something. up to you though.

source/kernel/src/forth.rs
```diff
@@ -191,6 +299,147 @@ impl Params {
...
+        if val.tid != tid {
+            return None;
+        }
+        let t = val.leaked.cast::<T>();
+        unsafe { Some(t.as_ref()) }
+    }
+}
+
+/// A container item for a type-erased object
+struct BohValue {
```
hawkw, line 421: this is kind of a homemade version of `Box<dyn Any>`, isn't it? if we could box unsized values, we could just use `Any` for this downcasting behavior...but we'd need to make substantial changes to the allocator for that, i guess.

hawkw (approved):
very nice! i left some annoying nits.

source/kernel/src/forth.rs
jamesmunns: Yeah, it would be needed if you wanted to do something like:

```rust
let t2 = stack.pop()?;
let t1 = stack.pop()?;
let something = boh.get_mut(t1)?;
let other = boh.get_mut(t2)?;
something.do_thing(other)?;
```

```forth
t1 t2 something::do_thing
```

source/kernel/src/forth.rs
jamesmunns: I am very open to replacing this with something less home-rolled! I just don't know how, and this is sorta how the "service discovery" table works too.

PR #50, opened 2023-06-02
update `mycelium` dependencies

### LINKED ISSUE #47: update to the latest maitake and cordyceps
the currently pinned dependencies are somewhat old

hawkw:
MnemOS' current dependencies on `maitake` and `cordyceps` from Mycelium
are somewhat old, and there have been a number of changes upstream. This
branch updates those dependencies to the latest Git commit upstream
(hawkw/mycelium@5149ee6146a9e0c1af56da79bb55578abf6dfddc). In addition,
`mnemos-alloc` is updated to
tosc-rs/mnemos-alloc@3b1b16d7c04f312a27befb1f1f7691f5f63cfb47.

Upstream changes in `maitake` have changed the `StaticScheduler` type's
task-spawning methods to require futures and their output to be `Send`,
in order to allow the type to implement `Send` and `Sync` soundly.
Therefore, because mnemOS currently spawns a large number of tasks which
cannot easily implement `Send`, this branch changes the kernel to use
the new `LocalStaticScheduler` type from `maitake` (added in
hawkw/mycelium#439), a `!Send` and `!Sync` scheduler which is capable of
spawning `!Send` futures

Closes #47

### DIFF
--- a/source/kernel/src/comms/bbq.rs
+++ b/source/kernel/src/comms/bbq.rs
@@ -13,7 +13,7 @@ use crate::fmt;
 use abi::bbqueue_ipc::{BBBuffer, Consumer as InnerConsumer, Producer as InnerProducer};
 use abi::bbqueue_ipc::{GrantR as InnerGrantR, GrantW as InnerGrantW};
 use maitake::sync::Mutex;
-use maitake::wait::WaitCell;
+use maitake::sync::WaitCell;
 use mnemos_alloc::{
     containers::{HeapArc, HeapArray},
     heap::AHeap,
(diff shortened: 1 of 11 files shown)


source/kernel/src/comms/bbq.rs
```diff
@@ -200,6 +200,9 @@ impl GrantR {
     }
 }
 
+unsafe impl Send for GrantR {}
```
jamesmunns, line 203: These aren't wrong but potentially not needed, do you want to keep them in, or back them out and add in a separate pass?

hawkw: i figured it was better to keep them, but :woman_shrugging:

PR #51, opened 2023-06-02
fix various clippy warnings

hawkw:
@jamesmunns it's up to you whether we care about this. i'm also happy to just disable clippy in my VSCode workspace for mnemOS...


jamesmunns (approved):
These all look fine to me!

PR #54, opened 2023-06-04
Fold the allocator back into the mnemos repo

jamesmunns:
This undoes a decision from about a year ago to move the allocator into its own crate.

Merged as an "unrelated history", meaning all history is retained.

This should NOT be a squash merge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1365,7 +1365,6 @@ dependencies = [
 [[package]]
 name = "mnemos-alloc"
 version = "0.1.0"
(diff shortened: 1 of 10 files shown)


hawkw (approved):
let's goooooo

hawkw:
i love how the first commit in this history is "[Export from mnemos super repo](https://github.com/tosc-rs/mnemos/pull/54/commits/6b7a00ffd8ba8fd869ff84d5902db80d1fc44ce8)"

PR #55, opened 2023-06-04
Restore missing commit from mnemos-alloc repo.

jamesmunns:
This commit was missed when I merged from the upstream repo:

https://github.com/tosc-rs/mnemos-alloc/pull/9

commit 3b1b16d7c04f312a27befb1f1f7691f5f63cfb47
Author: Eliza Weisman <eliza@buoyant.io>
Date:   Thu Jun 1 08:15:12 2023 -0700

  add missing `Send` impl for `AHeap` (#9)

  Currently, any Future that awaits one of the AHeap allocate,
  allocate_arc, allocate_array_with, or allocate_fixed_vec methods
  is !Send. This is because these methods return a Future which
  contains a &self receiver, and the AHeap type itself is !Send.
  This is sad and not very great.

  This branch adds a missing Send impl to AHeap. This should be safe
  for the same reason that the existing unsafe impl Sync for AHeap is
  safe.

  * add test asserting heap futures are Send
  * add missing Send impl for AHeap

### DIFF
--- a/source/alloc/src/heap.rs
+++ b/source/alloc/src/heap.rs
@@ -31,6 +31,7 @@ pub struct AHeap {
 // atomic operations to ensure the data is initialized and exclusively
 // accessed.
 unsafe impl Sync for AHeap {}
+unsafe impl Send for AHeap {}
 
 impl AHeap {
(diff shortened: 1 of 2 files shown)


hawkw (approved):
shipit

PR #57, opened 2023-06-04
Monorepo the `forth3` project

jamesmunns:
Keeps `forth3` and `f3repl`, `forthboy`'s parts now live in https://github.com/tosc-rs/teletype/.

Same deal as #54.

This should NOT be a squash merge.


source/forth3/.github/workflows/ci.yml
hawkw, line 1: we'll presumably need to delete this and roll it into the main CI config

source/forth3/.github/workflows/ci.yml
```diff
@@ -0,0 +1,165 @@
+name: CI
```
jamesmunns, line 1: This should eventually be integrated into mnemos' CI. It's pretty important IMO to preserve regular + miri testing.

hawkw (approved):
we should probably remove the separate CI config in forth3, because, AFAIK, a `.github` dir that isn't in the repo root doesn't do anything. other than that, LGTM!

source/forth3/.github/workflows/ci.yml
jamesmunns: Opened as #58!

PR #60, opened 2023-06-04
initial `maitake::time` integration

### LINKED ISSUE #46: maitake timer wheel integration
`maitake` includes a [hierarchical timer wheel implementation](https://mycelium.elizas.website/maitake/time/index.html). this is a data structure that's optimized for tracking timers of varying durations.

replacing the current timer implementation in mnemOS with `maitake::time` would let the kernel manage larger numbers of timers more efficiently.

hawkw:
This is a first pass at an integration with `maitake::time`'s timer
wheel.

I've added a `maitake` timer to the `Kernel`, changed the existing code
in `melpomene` that uses `Delay` to use the kernel's timer instead, and
added a new `sleep::ms` async builtin word to the Forth VMs that makes
the forth task sleep for that many milliseconds.

[Screencast from 2023-06-04 10-36-16.webm](https://github.com/tosc-rs/mnemos/assets/2796466/694226f1-7266-427d-8eac-2b5310c419f1)

Note that we can't (currently) simulate the hardware timer in oneshot
mode in melpomene. This is because we don't have a way to wait for any
"interrupt", since we... don't actually get interrupts in the simulator.
Right now the simulator uses a "timestamp-based" timer mode but it could
also simulate a periodic timer by spawning a separate Tokio task to
advance the timer. We *could* also do wait-for-interrupt behavior if we
added a channel to the sim drivers so they can notify the run loop on a
pretend interrupt, but I didn't want to implement that as part of this
branch.

Closes #46


Cargo.toml
```diff
@@ -17,7 +17,8 @@ members = [
 
 [patch.crates-io.maitake]
 git = "https://github.com/hawkw/mycelium.git"
-rev = "5149ee6146a9e0c1af56da79bb55578abf6dfddc"
+# rev = "5149ee6146a9e0c1af56da79bb55578abf6dfddc"
+branch = "eliza/next-deadline"
```
jamesmunns, line 21: Reminder to rev after the mycelium PR lands

source/kernel/src/forth.rs
```diff
@@ -435,6 +438,21 @@ async fn spawn_forth_task(forth: &mut forth3::Forth<MnemosContext>) -> Result<()
...
 }
 
+/// Binding for [`Kernel::sleep()`]
+///
+/// Sleep for the provided number of milliseconds.
+///
+/// Call: `MILLIS sleep`.
+/// Return: No change
+async fn sleep_ms(forth: &mut forth3::Forth<MnemosContext>) -> Result<(), forth3::Error> {
+    let millis = forth.data_stack.try_pop()?.as_i32();
```
jamesmunns, line 448: You probably want to check this is not negative, otherwise we're gunna be sleeping for a hot minute

source/kernel/src/lib.rs
```diff
@@ -240,11 +253,20 @@ impl Kernel {
...
+        inner.scheduler.tick()
 
         // TODO: Send time to userspace?
     }
 
+    /// Initialize the kernel's `maitake` timer as the global default timer.
+    ///
+    /// This allows the use of `sleep` and `timeout` free functions.
+    /// TODO(eliza): can the kernel just "do this" once it becomes active? Or,
+    /// have a "kernel.init()" or something that does this and other global inits?
```
jamesmunns, line 265: ATM, the only "init" is what we do in `new`. There are a couple steps that are "place the global thing", then use it.

We could document that `Kernel::new()` should only be called once.

FWIW: I haven't really been using any globals, and we might want to consider just having a kernel global that gives access to everything.

source/melpomene/src/main.rs
```diff
@@ -347,13 +348,29 @@ fn kernel_entry(opts: MelpomeneOptions) {
...
+        let tick = k.tick();
 
         KERNEL_LOCK.store(false, Ordering::Release);
+
+        // advance the timer
+        let ticks = t0.elapsed().as_millis() as u64;
+        let turn = k.timer().force_advance_ticks(ticks);
+        t0 = tokio::time::Instant::now();
+
+        if turn.expired == 0 && !tick.has_remaining {
```
jamesmunns, line 366: This seems generally acceptable to me, I'm not sure how we'd abstract that across other hardware platforms, but its okay if you don't solve "why are all these OS bits in the simulator and not the kernel" in this PR.

jamesmunns (approved):
Couple of comments, overall looks good!

Cargo.toml
hawkw: oh, yeah, this can be pointed back to main

source/kernel/src/forth.rs
hawkw: good call

source/kernel/src/lib.rs
hawkw: I would love to have a kernel global eventually --- I'm thinking this would be rolled into whatever sets up the global kernel context, when that exists. I don't think `Kernel::new` can do this, because the `set_global_timer` interface is only callable with a `&'static Timer`...

source/melpomene/src/main.rs
hawkw: yeah...the fact that this needs to actually figure out how much time has elapsed makes this a little more complex than just taking a "wait for interrupt" function or whatever. we'll need some kind of abstraction which i was thinking was better to figure out later.

PR #61, opened 2023-06-05
feat(melpo): fake "wait for interrupt"

hawkw:
This branch changes Melpomene's kernel main loop to behave somewhat more
like a "real" kernel on hardware would, by waiting for a (simulated)
interrupt or timer when the kernel executor has nothing else to do.
Since the simulator does not handle real hardware interrupts, we fake an
"interrupt" using a `tokio::sync::Notify` that the simulated kernel task
awaits when it is waiting for an interrupt. The simulated drivers wake
this `Notify` when they receive a "hardware interrupt".

### DIFF
--- a/source/melpomene/Cargo.toml
+++ b/source/melpomene/Cargo.toml
@@ -51,7 +51,7 @@ optional = true
 
 [dependencies.tokio]
 version = "1.19"
-features = ["rt", "time", "macros"]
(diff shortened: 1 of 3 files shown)


jamesmunns (approved):
Looks great to me. I will likely have more opinions once I start porting to the D1 and dealing with real interrupts.

PR #63, opened 2023-06-07
chore: run tests on CI

### LINKED ISSUE #58: [chore]: Restore forth3's CI
This should eventually be integrated into mnemos' CI. It's pretty important IMO to preserve regular + miri testing.

_Originally posted by @jamesmunns in https://github.com/tosc-rs/mnemos/pull/57#discussion_r1216069225_

hawkw:
Currently, CI just runs `cargo check` and `cargo fmt`, and never actually runs host tests. Some crates in the repo, such as `forth3`, contain tests that should be run on the host.

This branch adds a `cargo test` job to the CI workflow.

Closes #58


jamesmunns:
Looks like the build is failing because e-g-sim requires `libsdl2-dev` installed.

Would it make sense to restore the forth3 miri tests here too?

hawkw:
> Looks like the build is failing because e-g-sim requires `libsdl2-dev` installed.
> 
> Would it make sense to restore the forth3 miri tests here too?

yeah, it would!

PR #66, opened 2023-06-10
Actually no-std again

jamesmunns:
Fix some no-std issues with spitebuf, and support tracing-01 or tracing-02

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,6 @@ jobs:
       run: |
         cargo check \
           --workspace \
-          --all-features \
           --quiet \
           --message-format=json |
           cargo-action-fmt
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -73,8 +72,7 @@ jobs:
     - name: cargo nextest run
       run: |
         cargo nextest run \
-          --workspace \
-          --all-features
```
hawkw, line 77: does removing this mean that the forth3 tests aren't run with async, floats, etc enabled? because if that's the case, i would prefer an approach that lets us keep the `--all-features` build.

jamesmunns: The miri tests do still set this, but the normal tests don't. I'll rework this PR and add a mnemos-tracing crate instead.

hawkw (approved):
thanks i hate everything about this 👍

PR #67, opened 2023-06-10
Add a whole crate for this

jamesmunns:


jamesmunns:
This leads to problems with the `instrument` macro, we might revisit this in the future.

PR #69, opened 2023-06-11
Add D1 platform (2023 edition)

jamesmunns:


hawkw:
nice

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    // k.initialize(async {}).unwrap();
+
+    loop {
+        let tick = k.tick();
+
+        if !tick.has_remaining {
+            let mut delay = riscv::delay::McycleDelay::new(1_080_000_000);
+            delay.delay_ms(1);
+            let _turn = k.timer().force_advance_ticks(1);
```
hawkw, line 28: this isn't *really* correct: we delay for 1ms every tick, but we don't consider how long it took to tick the kernel scheduler, which may vary substantially based on how many tasks are scheduled on that tick and how long those tasks take to poll. 

i'm fine with merging this as is, but eventually we'll probably want to use a hardware timestamp instead of delaying?

platforms/allwinner-d1/drivers/src/lib.rs
hawkw, line 1: i wonder if this will be a useful abstraction on other platforms too, eventually...

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -0,0 +1,55 @@
+#![no_std]
+#![no_main]
+
+use core::{panic::PanicInfo, time::Duration};
+use drivers::Ram;
+use kernel::{Kernel, KernelSettings};
+use embedded_hal::blocking::delay::DelayMs;
+
+const HEAP_SIZE: usize = 384 * 1024 * 1024;
```
hawkw, line 9: would be cool if, eventually, there was a nice way to keep this in sync with the linker script...we can figure that out later though

hawkw (commented):
you love to see it!!

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+                riscv::asm::wfi();
+            }
+            // Disable the timer interrupt in case that wasn't what woke us up
+            plic.deactivate(Interrupt::TIMER1).unwrap();
+            timer1.set_interrupt_en(false);
+            timer1.stop();
+
+            // Account for time slept
+            let elapsed = wfi_start.wrapping_sub(timer0.current_value());
+            let _turn = k.timer().force_advance_ticks(elapsed.into());
```
hawkw, line 105: we should advance the timer wheel on every iteration of the loop even when there *are* tasks remaining in the run queue, so i think this should be moved out of the if. otherwise, expiring timers will not be notified when the system is busy. 

we should probably move advancing the timer wheel to *before* the "wait for interrupt" phase, and only wait for an interrupt if the `Turn` struct indicates that no timers completed. in general i think the overall structure should probably look closer to the run loop in melpomene.

hawkw: UPDATE: oh, i had missed that there was already a call to advance the wheel before the if. never mind, this looks right!

.github/workflows/ci.yml
```diff
@@ -74,6 +94,7 @@ jobs:
       run: |
         cargo nextest run \
           --workspace \
+          --exclude lichee-rv \
```
hawkw, line 97: nit, take it or leave it: maybe worth a comment explaining why this is excluded?

.github/workflows/ci.yml
```diff
@@ -126,6 +147,7 @@ jobs:
           --message-format=json \
           --no-deps \
           --all-features \
+          --exclude lichee-rv \
```
hawkw, line 150: maybe worth a comment explaining why this is excluded?

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+    // Loopback on virtual port zero
+    k.initialize(async move {
+        let mut hdl = loop {
+            match SerialMuxHandle::from_registry(k).await {
+                Some(c) => break c,
+                None => {
+                    k.sleep(Duration::from_millis(100)).await;
+                }
+            }
+        };
```
hawkw, line 114: nit/TIOLI: this pattern occurs a couple times, we could maybe make a helper for it?

alternatively, we could have the `k.initialize()` call that initializes the serial mux return a joinhandle or something that these init tasks can await, so that we don't need to sleep in a loop? not a blocker in either case, we can make this nicer later.

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
hawkw, line 1: how much of the code in this file is Lichee-RV specific, and how much of it can be shared across other D1 boards? could be nice to (eventually) split out stuff that's SOC-specific, rather than board-specific?

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+
+        // If there is nothing else scheduled, and we didn't just wake something up,
+        // sleep for some amount of time
+        if !tick.has_remaining && turn.expired != 0 {
+            let wfi_start = timer0.current_value();
+
+            // TODO(AJM): Sometimes there is no "next" in the timer wheel, even though there should
+            // be. Don't take lack of timer wheel presence as the ONLY heuristic of whether we
+            // should just wait for SOME interrupt to occur. For now, force a max sleep of 100ms
+            // which is still probably wrong.
```
hawkw, line 188: why is it probably wrong?

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+fn initialize_kernel() -> Result<&'static Kernel, ()> {
+    let k_settings = KernelSettings {
+        heap_start: AHEAP.as_ptr(),
+        heap_size: HEAP_SIZE,
+        max_drivers: 16,
+        k2u_size: 4096,
+        u2k_size: 4096,
+        timer_granularity: Duration::from_nanos(333),
```
hawkw, line 234: where does 333 comef from? could maybe be worth a comment?

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+    };
+    let k = unsafe { Kernel::new(k_settings).map_err(drop)?.leak().as_ref() };
+    Ok(k)
+}
+
+#[panic_handler]
+fn handler(info: &PanicInfo) -> ! {
```
hawkw, line 241: would eventually be nice if the kernel could provide a cross platform panic hook with nice formatting, but that should be a later thing...

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+    // Ugly but works
+    let mut uart: Uart = unsafe { core::mem::transmute(()) };
```
hawkw, line 243: what the fuck

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    write!(&mut uart, "\r\n").ok();
+    write!(&mut uart, "{}\r\n", info).ok();
+
+    loop {
+        core::sync::atomic::fence(Ordering::SeqCst);
+    }
+}
+
+fn handle_dmac() {
```
hawkw, line 253: this is the DMAC ISR, right? maybe worth a comment stating that?

platforms/allwinner-d1/boards/lichee-rv/src/uart.rs
hawkw, line 1: is this Lichee-RV specific, or is this UART built into the D1 SOC?

platforms/allwinner-d1/boards/lichee-rv/src/uart.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    pub fn handle_uart0_int() {
+        let uart0 = unsafe { &*UART0::PTR };
+        let prod = UART_RX.load(Ordering::Acquire);
+        if !prod.is_null() {
+            let prod = unsafe { &*prod };
+
+            while let Some(mut wgr) = prod.send_grant_max_sync(64) {
+                let used_res = wgr.iter_mut().enumerate().try_for_each(|(i, b)| {
+                    if uart0.usr.read().rfne().bit_is_set() {
```
hawkw, line 66: might be nice if, somewhere, there was a comment saying what RFNE and RBR are?

platforms/allwinner-d1/boards/lichee-rv/src/uart.rs
```diff
@@ -0,0 +1,181 @@
...
+                    Ok(()) => {
+                        let len = wgr.len();
+                        wgr.commit(len);
+                    }
+                    Err(used) => {
+                        wgr.commit(used);
+                        break;
+                    }
+                }
+            }
```
hawkw, line 84: nit, TIOLI: this logic kinda feels like it could maybe be simplified with a for loop:
```suggestion
            'read: while let Some(mut wgr) = prod.send_grant_max_sync(64) {
                for (used, b) in wgr.iter_mut().enumerate() {
                    if !uart0.usr.read().rfne().bit_is_set() {
                        wgr.commit(used);
                        break 'read;
                    }

                    *b = uart0.rbr().read().rbr().bits();
                }

                let len = wgr.len();
                wgr.commit(len);
            }
```
which, AFAICT, should do the same thing?

platforms/allwinner-d1/boards/lichee-rv/src/uart.rs
```diff
@@ -0,0 +1,181 @@
...
+        let (fifo_a, fifo_b) = new_bidi_channel(k.heap(), cap_in, cap_out).await;
+
+        let _server_hdl = k.spawn(D1Uart::serial_server(fifo_b, kcons)).await;
+
+        let (prod, cons) = fifo_a.split();
+        let _send_hdl = k.spawn(D1Uart::sending(cons, tx_channel)).await;
+
+        let boxed_prod = k.heap().allocate(prod).await;
+        let leaked_prod = boxed_prod.leak();
+        UART_RX.store(leaked_prod.as_ptr(), Ordering::Release);
```
hawkw, line 173: nit: should we maybe do a swap/compare-exchange so we can assert the previous value isn't null?

platforms/allwinner-d1/drivers/src/dmac/descriptor.rs
```diff
@@ -0,0 +1,299 @@
...
+pub struct Descriptor {
+    configuration: u32,
+    source_address: u32,
+    destination_address: u32,
+    byte_counter: u32,
+    parameter: u32,
+    link: u32,
+}
+
+// TODO: THIS COULD PROBABLY BE A BITFIELD LIBRARY
```
hawkw, line 12: i might make a PR for this later

platforms/allwinner-d1/drivers/src/plic.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub fn new(plic: PLIC) -> Self {
+        // TODO any initial setup we should be doing for the PLIC at startup?
+        Self { plic }
+    }
+
+    /// Obtain a static `Plic` instance for use in e.g. interrupt handlers
+    ///
+    /// # Safety
+    ///
+    /// 'Tis thine responsibility, that which thou doth summon.
```
hawkw, line 31: "don't summon anything bigger than your head" :D

platforms/allwinner-d1/drivers/src/plic.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    pub fn complete(&self, interrupt: Interrupt) {
+        self.plic
+            .mclaim
+            .write(|w| w.mclaim().variant(interrupt.into_bits() as u16));
+    }
+
+    pub unsafe fn register(&self, interrupt: Interrupt, new_hdl: fn()) {
+        let v = INTERRUPT_ARRAY.iter().find(|v| v.id == interrupt as u16);
```
hawkw, line 83: are the IDs numerically increasing, and if so, can we just put each interrupt at its' ID's index in the array, instead of doing this? not a blocker, just curious.

platforms/allwinner-d1/drivers/src/plic.rs
```diff
@@ -0,0 +1,287 @@
...
+        let nr = interrupt.into_bits() as usize;
+        self.plic.prio[nr].write(|w| w.bits(priority.into_bits()));
+    }
+
+    pub fn claim(&self) -> Interrupt {
+        let claim = self.plic.mclaim.read().mclaim().bits() as u8;
+        match Interrupt::try_from(claim) {
+            Ok(interrupt) => interrupt,
+            Err(_) => {
+                panic!("error claiming interrupt");
```
hawkw, line 71: nit: should the panic message include the error?

tools/crowtty/src/main.rs
```diff
@@ -12,40 +13,83 @@ pub struct Chunk {
...
+            .unwrap();
+        Connect::Serial(port)
+    }
+}
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let args = std::env::args().skip(1).collect::<Vec<_>>();
+    let args = args.iter().map(String::as_str).collect::<Vec<_>>();
+
+    let mut port = match args.as_slice() {
```
hawkw, line 67: nit: at some point it might be nice to just use `clap` for arg parsing here --- we already use it in `melpomene`. not a blocker for this PR though...

hawkw (approved):
Some nits, but this is awesome!

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
jamesmunns: It's a ZST, so this works. I probably should just have an unsafe "summon" method (like `Plic` has), but... yeah.

jamesmunns:
@hawkw breaking this out as a longer response:

> is this Lichee-RV specific, or is this UART built into the D1 SOC?

So, some of these things COULD be portable to any user of the D1, however there's a very fuzzy line today. I haven't really tackled any of the "portability challenges", like handling different configs, or pin mapping, or what DMA channel to use, or even how to handle interrupts.

These absolutely *should* live in the `allwinner-d1/drivers` crate, but the current state of the code isn't suitable for that yet. I could put it there, but it's all very hardcoded for just this one board, so far.

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
jamesmunns: The timer is 3MHz, so the granularity is 1/3 of a microsecond. Will add a comment.

platforms/allwinner-d1/drivers/src/plic.rs
jamesmunns: We could, we would just need a sparse array, as some interrupt IDs are not used.

I think there is SOME way to get the hardware to do normal ISR dispatching, instead of a single "hardware interrupt method", but I haven't looked up how to do that yet.

We also could probably binary search, but I decided to go with 'bad and simple and done' for now.

tools/crowtty/src/main.rs
jamesmunns: Agreed! Happy to swap to that in the future esp if we have more options.

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
jamesmunns: Blindly waiting 100ms is not *great*, it keeps us from sleeping *forever*, which is something I saw during testing (it may have been because I wasn't previously checking the `expired` field of the timer wheel turn), but it also means that we could potentially sleep for longer than we should have.

platforms/allwinner-d1/boards/lichee-rv/src/uart.rs
jamesmunns: In this case, the previous value SHOULD be null.

PR #72, opened 2023-06-18
platform(lichee-rv): Re-add hacky SPI driver

jamesmunns:
This is enough to get the display rendering a pattern again.

This doesn't integrate with the EmbDisplay stuff yet, but is enough to actively draw to the LCD.

### DIFF
--- a/platforms/allwinner-d1/boards/lichee-rv/src/main.rs
+++ b/platforms/allwinner-d1/boards/lichee-rv/src/main.rs
@@ -14,7 +14,7 @@ use kernel::{
     drivers::serial_mux::{RegistrationError, SerialMuxClient, SerialMuxServer},
     Kernel, KernelSettings,
 };
-use spim::{kernel_spim1, SpiSender, SPI1_TX_DONE};
+use spim::{kernel_spim1, SpiSenderClient, SpiSenderServer, SPI1_TX_DONE};
 use uart::D1Uart;
 mod spim;
 mod uart;
@@ -73,19 +73,23 @@ fn main() -> ! {
     let dmac = Dmac::new(p.DMAC, &mut p.CCU);
     let [ch0, ..] = dmac.channels;
     dmac.dmac.dmac_irq_en0.modify(|_r, w| {
+        // used for UART0 DMA sending
         w.dma0_queue_irq_en().enabled();
+        // used for SPI1 DMA sending
+        w.dma1_queue_irq_en().enabled();
         w
     });
 
     // Initialize SPI stuff
     k.initialize(async move {
-        SpiSender::register(k, 4).await.unwrap();
+        // Register a new SpiSenderServer
+        SpiSenderServer::register(k, 4).await.unwrap();
     })
     .unwrap();
 
     k.initialize(async move {
         k.sleep(Duration::from_millis(100)).await;
-        let mut spim = SpiSender::from_registry(k).await.unwrap();
+        let mut spim = SpiSenderClient::from_registry(k).await.unwrap();
         // println!("GOT SPIM");
         // SPI_BCC (0:23 and 24:27)
         // SPI_MTC and SPI_MBC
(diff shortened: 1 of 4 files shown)


platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -74,6 +77,71 @@ fn main() -> ! {
         w
     });
 
+    // Initialize SPI stuff
+    k.initialize(async move {
+        SpiSender::register(k, 4).await.unwrap();
```
hawkw, line 82: what does `4` mean here?

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -74,6 +77,71 @@ fn main() -> ! {
...
+    })
+    .unwrap();
+
+    k.initialize(async move {
+        k.sleep(Duration::from_millis(100)).await;
+        let mut spim = SpiSender::from_registry(k).await.unwrap();
```
hawkw, line 88: it would be nice if `SpiSender::register` could return a handle --- that way, we could await the `JoinHandle` of the first `k.initialize(...)` call in the second one, instead of having to sleep and then try to get it from the registry. Just a thought.

platforms/allwinner-d1/boards/lichee-rv/src/spim.rs
hawkw, line 1: is this Lichee RV specific, or can it be used on other D1 boards?

platforms/allwinner-d1/boards/lichee-rv/src/spim.rs
```diff
@@ -0,0 +1,224 @@
...
+                        w
+                    });
+                    spi.spi_mbc.modify(|_r, w| {
+                        w.mbc().variant(payload.len() as u32);
+                        w
+                    });
+                    spi.spi_mtc.modify(|_r, w| {
+                        w.mwtc().variant(payload.len() as u32);
+                        w
+                    });
```
hawkw, line 112: it could be nice if there were comments here saying what these different registers do?

platforms/allwinner-d1/boards/lichee-rv/src/spim.rs
```diff
@@ -0,0 +1,224 @@
...
+                    let descriptor = d_cfg.try_into().map_err(drop)?;
+                    unsafe {
+                        let mut chan = Channel::summon_channel(1);
+                        chan.set_channel_modes(ChannelMode::Wait, ChannelMode::Handshake);
+                        chan.start_descriptor(NonNull::from(&descriptor));
+                    }
+                    match SPI1_TX_DONE.wait().await {
+                        Ok(_) => {}
+                        Err(_) => todo!(),
+                    }
```
hawkw, line 144: nit, take it or leave it: this could probably just be
```suggestion
                    SPI1_TX_DONE
                        .wait()
                        .expect("SPI1_TX_DONE WaitCell should never be closed");
```
since waiting on a `WaitCell` only errors if we have called `WaitCell::close`, and AFAIK we never do that here.

hawkw (approved):
looks good to me as a first pass --- i left a few minor comments but feel free to ignore them for now!

platforms/allwinner-d1/boards/lichee-rv/src/spim.rs
```diff
@@ -0,0 +1,224 @@
...
+        w.pd11_select().spi1_clk_dbi_sclk();
+        w.pd12_select().spi1_mosi_dbi_sdo();
+        w
+    });
+    gpio.pd_pull0.write(|w| {
+        w.pd10_pull().pull_disable();
+        w.pd11_pull().pull_disable();
+        w.pd12_pull().pull_disable();
+        w
+    });
```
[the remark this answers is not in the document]
jamesmunns: Yup

platforms/allwinner-d1/boards/lichee-rv/src/spim.rs
jamesmunns: Same deal as last time, it *could* be made generic, but isn't yet.

platforms/allwinner-d1/boards/lichee-rv/src/spim.rs
jamesmunns: I think you missed an `await`, but I agree with you

PR #73, opened 2023-06-19
Handle multiple lines in forth output

jamesmunns:
Before this, outputting newlines in forth would skip them.


hawkw (approved):
looks good to me!

Issue #74, opened 2023-06-19
Design Topics around services + messaging


jamesmunns:
So here are a couple things I am realizing as I am working on services and stuff. This should probably be a couple issues/action items, but I am dumping my thoughts here instead of in chat.

## We should think about our use of `alloc` in kernel services

Originally, I loosely had the plan that "services can alloc when starting, but shouldn't alloc later". This is why you see all the "how many clients/messages/frames does this service take" in `register()` calls.

On one hand: it'd probably be good for smaller systems to have bounded memory usage, more or less just acting like a bump allocator on startup, and not requiring variable runtime allocation.

On the other hand: It's sometimes a pain in the butt to actually do this/very convenient to alloc whenever.

I dunno if there is a "right" answer, but picking one and being consistent is probably a good thing.

## Messaging should be easier/better encapsulated

There's a LOT of impl details exposed to every service. The core interface is:

* Servers receive messages (from many clients)
* Servers optionally reply to messages

There is SOME shenanigans whether the message types are `Serialize` or not, that limits whether a server can advertise ONLY to other kernel client entities, or (in the future) userspace or proxied/remote client entities, but right now there are a ton of `Message`/`Envelope` whatever whatever types that are kinda gross to actually use.

This could probably use some better encapsulation, and nicer APIs on both the server and client side.

## Service definition should be more consistent

I'm sort of learning how services "should be" as we are writing them.

I think at the moment, my best practice guide is:

* Services should be defined (whenever possible) in the kernel
* Services should provide a reasonable client implementation
* `RegisteredDriver` should be implemented on a "nothing" type, that can be used to get a client `Nothing::from_registry() -> NothingClient` or so.
* Services defined in the kernel SHOULD NOT provide a Server implementation, unless they only rely on other services (like `SerMux` does).

hawkw:
> ## We should think about our use of `alloc` in kernel services

Another "lesson learned" for me is that, with the async `mnemos-alloc` API, the failure mode for OOM conditions is "a task trying to allocate memory waits forever". When some allocations are dynamic and short-lived, this isn't terrible, as that memory may later be released and the task can just wait until its available. However, if the allocation occurs while *initializing* a kernel task that lives for the entire runtime of the OS, such as a driver service, and the only other existing allocations are for similarly long-running driver services, the whole system just gets completely stuck. And, because the allocator doesn't have any logging or similar diagnostics, there's no way to determine *why* the system is stuck.

[...]

PR #75, opened 2023-06-19
Rework services

jamesmunns:
This PR achieves two things:

1. It moves the `EmbDisplay` service definition back into the Kernel, so I can implement it for the D1
2. It does a "cleanup pass" on the existing services to be more consistent. Context: #74.

For now, I've picked up the pattern of:

* a **Service**, which is a ZST that implements the `RegisteredDriver` trait
* a **Client**, what you use to talk to the service
* a **Server**, which implements the service

I've also added some block comments to make things easier to look at.


source/kernel/src/drivers/emb_display.rs
```diff
@@ -0,0 +1,212 @@
...
+            .await?;
+
+        Some(EmbDisplayClient {
+            prod,
+            reply: Reusable::new_async(kernel).await,
+        })
+    }
+
+    /// Drop the provided framechunk without drawing
+    pub async fn drop_framechunk(&mut self, chunk: FrameChunk) -> Result<(), ()> {
```
hawkw, line 114: it seems like this should maybe return `Result<(), FrameError>`?

source/kernel/src/drivers/serial_mux.rs
```diff
@@ -5,33 +5,31 @@ use crate::{
...
 };
 use maitake::sync::Mutex;
 use mnemos_alloc::containers::{HeapArc, HeapArray, HeapFixedVec};
 use uuid::Uuid;
 
+////////////////////////////////////////////////////////////////////////////////
+// Service Definition
+////////////////////////////////////////////////////////////////////////////////
+
 /// SerialMux is the registered driver type
```
hawkw, line 20: maybe worth a comment explaining that this module implements the serial mux server as well as the client and registered driver?

source/kernel/src/drivers/emb_display.rs
```diff
@@ -0,0 +1,212 @@
...
+    prelude::*,
+};
+use mnemos_alloc::containers::HeapArray;
+use uuid::Uuid;
+
+////////////////////////////////////////////////////////////////////////////////
+// Service Definition
+////////////////////////////////////////////////////////////////////////////////
+
+// Registered driver
```
hawkw, line 39: nit: maybe worth including a doc comment for the services which expect to be implemented by platform impls explaining that there's no actual server task implemented in the kernel lib? like
```suggestion
/// Registered driver type for the `EmbDisplay` service.
///
/// This module provides an implementation of the client for this service, but
/// not the server. A server implementing this service must be provided by the
/// hardware platform implementation.
```

source/kernel/src/drivers/simple_serial.rs
```diff
@@ -0,0 +1,74 @@
...
+        })
+    }
+
+    pub async fn get_port(&mut self) -> Option<BidiHandle> {
+        self.kprod
+            .send(
+                Request::GetPort,
+                ReplyTo::OneShot(self.rosc.sender().await.ok()?),
+            )
+            .await
```
hawkw, line 67: it occurs to me that a a pattern like this shows up kind of a lot, maybe it could be factored out into a helper fn (in another PR, probably)?

source/melpomene/src/sim_drivers/emb_display.rs
```diff
@@ -31,31 +31,15 @@ use embedded_graphics_simulator::{
...
-pub struct EmbDisplay;
-
-// impl EmbDisplay
-impl RegisteredDriver for EmbDisplay {
-    type Request = Request;
-    type Response = Response;
-    type Error = FrameError;
-    const UUID: Uuid = mnemos_kernel::registry::known_uuids::kernel::EMB_DISPLAY;
-}
+pub struct SimDisplay;
```
hawkw, line 40: maybe worth a comment like
```suggestion
/// Implements the [`EmbDisplayService`] driver using the `embedded-graphics`
/// simulator.
pub struct SimDisplay;
```

PR #76, opened 2023-06-20
kernel: add helper fn for oneshot driver requests

hawkw:
Depends on #75 

This commit adds a `KHandle::request_oneshot` helper function that bundles together code for sending a request with a reusable oneshot sender as the message, and awaiting the response. This pattern occurred a few times in service client implementations, so I thought it might be nice to have a helper method for it.

We can add similar methods for other types of requests later.


jamesmunns (approved):
Looks great, thank you :)

PR #77, opened 2023-06-20
loltimer?

hawkw:


jamesmunns:
As noted in chat, this branch makes things... differently bad.

Before this change: 

2023-06-20-190234_2560x1440_scrot

After this change:

2023-06-20-190639_2560x1440_scrot

Both graphs should be a 50% duty cycle with a 200ms period (100 on, 100 off).

jamesmunns:
Closing this as it will be addressed in #78

PR #79, opened 2023-06-21
chore: add a Justfile to automate D1 flashing

hawkw:
Currently, flashing an Allwinner D1 board with mnemOS requires a bunch
of manual steps. This branch adds a quick Justfile with `build-d1` and
`flash-d1` recipes.

We may want to improve the build tooling more later, but this works for
now.

### DIFF
--- a/tools/_util.sh
+++ b/tools/_util.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+# utility functions used in other shell scripts.
+#
+# currently, this includes:
+# - cargo-style stderr logging (`err`, `note`, and `status` functions)
+# - confirmation prompts (`confirm` function)
+set -euo pipefail
+
+# Log an error to stderr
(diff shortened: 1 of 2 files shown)


tools/_util.sh
```diff
@@ -0,0 +1,80 @@
...
+# Returns the path to a Mycelium crate.
+#
+# Args:
+#     $1: crate name
+#
+# Returns:
+#     0 if the crate exists, 0 if it does not exist.
+crate_path() {
+    local crate="$1"
+    local mycoprefix='mycelium-';
```
jamesmunns, line 71: not against keeping this, but it might not work :)

hawkw: whoops i copied this file from mycelium and didn't remove anything lol

jamesmunns (approved):
Hell yeah

PR #80, opened 2023-06-22
feat(trace): add UART tracing

hawkw:
This branch adds an initial implementation of a `tracing` collector that
writes traces to virtual serial port 3 in a COBS+`postcard` binary
format. In addition, it adds a trace viewer capability to `crowtty` when
demuxing port 3, so that the traces written to port 3 are displayed in
`crowtty`'s output. The previous output from `crowtty` that records
every chunk of bytes read/written is now only displayed when a
`--verbose` argument is passed.

Normal `crowtty` output: 

image

Verbose mode:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,6 +60,55 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.3.2"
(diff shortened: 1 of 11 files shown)


hawkw:
image

PR #86, opened 2023-06-23
chore: split `allwinner-d1/boards` into its own workspace 

hawkw:
Currently, any time the D1 kernel binary is built, all Cargo caches for
crates built targeting the host are invalidated. Similarly, any time a
crate is built for the host target (e.g. to run `crowtty`), everything
built for the board is invalidated. This is an unfortunate limitation of
Cargo workspaces that contain crates built for different targets, and
means that a development workflow that invovles flashing the D1 and
running `crowtty` to talk to it involves an uncomfortable amount of
recompiling.

In order to fix this, this PR splits the `platforms/allwinner-d1/boards`
directory out of the main workspace, into a new workspace. This way, we
can have two separate cargo caches, one for the host, and one for the
board target platform. We can do a similar maneuver when adding support
for other targets.

This does have the tradeoff that running `cargo {check, build, fmt,
test} --workspace` will no longer build the D1 platform impl. However,
I've left the `drivers` crate as part of the main workspace, so that the
drivers are checked when checking everything else, and I've added new
Just recipes `test`, `check`, and `fmt`, which ensure that the D1
platform impl is also tested/built/checked/formatted.


jamesmunns:
Weird that https://github.com/tosc-rs/mnemos/actions/runs/5359393688/jobs/9722925553?pr=86#step:6:595 didn't fire before, afaict you didn't change anything here.

jamesmunns (approved):
Overall lgtm, +1 to merge after CI is green.

hawkw:
> Weird that https://github.com/tosc-rs/mnemos/actions/runs/5359393688/jobs/9722925553?pr=86#step:6:595 didn't fire before, afaict you didn't change anything here.

yeah idk what's up with that, but i'll fix it here anyway!

PR #91, opened 2023-06-24
feat(crowtty): logs indicate if target is TCP/UART

hawkw:
This commit adds a small additional indicate to `crowtty`'s logs to show whether the target connection is over TCP serial or over the UART.


jamesmunns (approved):
Cool! We might want CLI or config control eventually, but not today is okay.

PR #92, opened 2023-06-24
feat(trace): host-controlled tracing

hawkw:
Currently, the `tracing` collector in the Allwinner D1 platform impl is
hardcoded to the DEBUG level. This is unfortunate, as it means that we
will always do the work of collecting and formatting DEBUG traces, even
when no one is listening on the UART tracing port. It would be more
efficient if we only formatted traces when a `crowtty` instance has
subscribed to the UART tracing port. Additionally, because trace
metadata is sent when the callsite is initially registered, a `crowtty`
instance that attaches after the board has started running may miss some
trace metadata. Finally, there is no way to change the trace level on
the fly, as it's hard-coded in the kernel.

This branch changes the `mnemos-trace-proto` wire protocol to include a
host to target message to select a `tracing` level. Now, the board can
start up with all tracing disabled, and `crowtty` can send the desired
tracing level once it connects. This way, we don't collect or format
traces at all when `crowtty` isn't connected. In order to ensure
`crowtty` waits to send the tracing level message only once the board's
tracing code has been initialized, we add a periodic "heartbeat" message
sent by the board on the tracing port while tracing is idle. This is
used by the host which is listening for traces to detect the presence of
the tracing collector on the debug target. The same message is used to
ack a successful request to select the tracing level.

When the tracing level is selected, the target rebuilds `tracing`'s
callsite cache, which means all metadata for callsites which have
previously been hit will be sent to the host. This ensures that
`crowtty` always has the metadata for all tracing spans and events that
are enabled, even if the target encountered them before `crowtty`
connected.

Making this work required fixing an upstream `tracing` bug,
https://github.com/tokio-rs/tracing/pull/2634, where calling into code
that includes `tracing` diagnostics from inside a
`Collect::register_callsite` method would cause a deadlock.
Additionally, I had to add code to the collector for tracking whether
it's inside its own `send` method, in order to temporarily disable `bbq`
tracing. This is because the collector's use of `bbq` creates an
infinite loop when the TRACE level is enabled.

Check this out:
image

### DIFF
--- a/source/kernel/src/trace.rs
+++ b/source/kernel/src/trace.rs
@@ -35,7 +35,7 @@ impl SerialCollector {
     const CAPACITY: usize = 1024 * 4;
 
     pub const fn new() -> Self {
-        Self::with_max_level(LevelFilter::INFO)
+        Self::with_max_level(LevelFilter::OFF)
     }
 
     pub const fn with_max_level(max_level: LevelFilter) -> Self {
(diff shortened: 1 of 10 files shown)


hawkw:
In a subsequent branch, I intend to also add a host -> target heartbeat, so that the target can detect when `crowtty` has disconnected, and turn tracing off again.

source/kernel/src/trace.rs
```diff
@@ -25,7 +25,11 @@ pub struct SerialCollector {
     // TODO(eliza): Currently, this is recorded but not actually consumed...
     dropped_events: AtomicUsize,
 
-    max_level: LevelFilter,
+    max_level: AtomicU8,
+
+    /// Tracks whether we are inside of the collector's `send_event` method, so
```
jamesmunns, line 30: Love it

source/kernel/src/trace.rs
```diff
@@ -82,29 +99,116 @@ impl SerialCollector {
...
+        port: serial_mux::PortHandle,
+        k: &'static crate::Kernel,
+    ) {
+        use futures::FutureExt;
+        use maitake::time;
+        use postcard::accumulator::{CobsAccumulator, FeedResult};
+
+        // we probably won't use 256 whole bytes of cobs yet since all the host
+        // -> target messages are quite small
+        let mut cobs_buf: CobsAccumulator<16> = CobsAccumulator::new();
```
jamesmunns, line 120: We're gunna want some kind of config soon.

source/kernel/src/trace.rs
```diff
@@ -116,17 +220,28 @@ impl Collect for SerialCollector {
...
-            // if we couldn't send the metadata, skip this callsite, because the
-            // consumer will not be able to understand it without its metadata.
-            tracing_core_02::Interest::never()
+        // If we couldn't send the metadata, skip this callsite, because the
+        // consumer will not be able to understand it without its metadata.
+        if !sent {
+            return tracing_core_02::Interest::never();
+        }
+
+        // Due to the fact that the collector uses `bbq` internally, we must
```
jamesmunns, line 229: This *feels* a little icky, because this is a platform-specific impl detail - melpo doesn't use bbqueue for the tracing collection.

Might be worth instead to::

* Provide bbqueue methods that don't emit traces (and use them in lichee-rv)
* Start an "integration guide" that tells people to use something that doesn't emit traces when implementing tracing

I don't think this is a today blocker, but if we merge this as-is it might be good to throw up an issue to fix this someday.

source/kernel/src/trace.rs
```diff
@@ -195,3 +310,25 @@ impl Collect for SerialCollector {
...
+        LevelFilter::TRACE => 0,
+        LevelFilter::DEBUG => 1,
+        LevelFilter::INFO => 2,
+        LevelFilter::WARN => 3,
+        LevelFilter::ERROR => 4,
+        LevelFilter::OFF => 5,
+    }
+}
+
+const fn u8_to_level(level: u8) -> LevelFilter {
```
jamesmunns, line 325: It *might* be good to have an invalid level be an error (and ignore the "change log level command" instead), otherwise if we receive a garbage message that HAPPENS to look like a command, we have a 251/256 chance of just shutting off all logging. Tho I guess we don't really need to log at all to a connection that is sending us garbage. Hm.

source/trace-proto/src/lib.rs
```diff
@@ -2,11 +2,14 @@
 
 use core::{fmt, num::NonZeroU64};
 use tracing_serde_structured::{
-    SerializeId, SerializeMetadata, SerializeRecordFields, SerializeSpanFields,
+    SerializeId, SerializeLevel, SerializeMetadata, SerializeRecordFields, SerializeSpanFields,
 };
 
 #[derive(serde::Serialize, serde::Deserialize)]
 pub enum TraceEvent<'a> {
+    /// Sent by the target periodically when not actively tracing, to indicate
```
jamesmunns, line 10: We might want to mark this non-exhaustive and add things to the end instead, to avoid breaking wire format, once we care about that.

source/kernel/src/trace.rs
```diff
@@ -82,29 +99,116 @@ impl SerialCollector {
...
-            port.send(&rgr[..]).await;
+            'cobs: while !window.is_empty() {
+                window = match cobs_buf.feed_ref::<HostRequest>(window) {
+                    FeedResult::Consumed => break 'cobs,
+                    FeedResult::OverFull(new_wind) => new_wind,
+                    FeedResult::DeserError(new_wind) => new_wind,
+                    FeedResult::Success { data, remaining } => {
+                        match data {
+                            HostRequest::SetMaxLevel(lvl) => {
+                                let level = lvl.map(|lvl| lvl as u8).unwrap_or(5);
```
jamesmunns, line 132: Maybe note what "5" is here?

jamesmunns (approved):
Overall looks fine to me to merge! Left some comments, most are "some day maybe tweak this", take them or leave them :)

source/kernel/src/trace.rs
hawkw: definitely, yeah

source/kernel/src/trace.rs
hawkw: Well, *this* collector does use `bbqueue` --- so tif we are ever using the collector in this file, we are using bbqueue.

source/kernel/src/trace.rs
hawkw: these functions are used internally for converting the `AtomicU8` to/from a `LevelFilter`, not for parsing the log level sent to us from the host. that gets deserialized using `postcard`, so if we get an invalid value, we will error.

source/trace-proto/src/lib.rs
hawkw: yeah, i figured we didn't currently care about stability for the crate API _or_ for the wire format, but we can do that in the future.

source/kernel/src/trace.rs
hawkw: changed this to actually use the enum variant

PR #94, opened 2023-06-24
Reorganize Melpo Init

jamesmunns:
This moves the majority of melpo's init into smaller, separate service launches.

Additionally, it adds `register_no_retry` (and similar functions), with the existing `register` methods now doing the retry loop.

Additionally additionally, It moves the graphical forth shell into a new module in the kernel, usable by both hardware and the simulator.


source/kernel/src/drivers/serial_mux.rs
```diff
@@ -76,7 +78,28 @@ pub struct SerialMuxClient {
 }
 
 impl SerialMuxClient {
-    pub async fn from_registry(kernel: &'static Kernel) -> Option<Self> {
+    // Obtain a SerialMuxClient
+    //
+    // If the [SerialMuxServer] hasn't been registered yet, we will retry until it has been
```
hawkw, line 83:
```suggestion
    /// Obtain a `SerialMuxClient`
    ///
    /// If the [`SerialMuxServer`] hasn't been registered yet, we will retry until it has been
```

source/kernel/src/drivers/serial_mux.rs
```diff
@@ -76,7 +78,28 @@ pub struct SerialMuxClient {
...
+            }
+        }
+    }
+
+    // Obtain a SerialMuxClient
+    //
+    // Does NOT attempt to get a [SerialMuxServer] handle more than once.
+    //
+    // Prefer [SerialMuxClient::from_registry] unless you will not be spawning one
+    // around the same time as obtaining a client.
```
hawkw, line 101:
```suggestion
    /// Obtain a `SerialMuxClient`
    ///
    /// Does NOT attempt to get a [`SerialMuxServer`] handle more than once.
    ///
    /// Prefer [`SerialMuxClient::from_registry`] unless you will not be spawning one
    /// around the same time as obtaining a client.
```

source/kernel/src/drivers/serial_mux.rs
```diff
@@ -134,10 +169,39 @@ impl PortHandle {
 pub struct SerialMuxServer;
 
 impl SerialMuxServer {
+    // Register the SerialMuxServer.
+    //
+    // Will retry to obtain a [SimpleSerialClient] until success.
```
hawkw, line 174:
```suggestion
    /// Register the `SerialMuxServer`.
    ///
    /// Will retry to obtain a [`SimpleSerialClient`] until success.
```

source/kernel/src/drivers/serial_mux.rs
```diff
@@ -134,10 +169,39 @@ impl PortHandle {
...
+        }
+        Ok(())
+    }
+
+    // Register the SerialMuxServer.
+    //
+    // Does NOT attempt to obtain a [SimpleSerialClient] more than once.
+    //
+    // Prefer [SerialMuxServer::register] unless you will not be spawning one around
+    // the same time as registering this server.
```
hawkw, line 200:
```suggestion
    /// Register the SerialMuxServer.
    ///
    /// Does NOT attempt to obtain a [`SimpleSerialClient`] more than once.
    ///
    /// Prefer [`SerialMuxServer::register`] unless you will not be spawning one around
    /// the same time as registering this server.
```

hawkw (approved):

PR #96, opened 2023-06-25
Introduce a hybrid global/async allocator

jamesmunns:
This PR reworks the `mnemos-alloc` crate significantly, removing the
old "cursed but fancy" wrapper, and instead retains only some of the
basic "yield on OOM" behaviors.

This change was made to make it easier to use existing libraries and
collection types from the `alloc` crate, while also making it possible
to create failable allocation types by wrapping existing types, rather
than reimplementing them wholesale.


source/kernel/src/lib.rs
```diff
@@ -366,12 +261,12 @@ impl Kernel {
         f(&mut guard)
     }
 
-    pub fn spawn_allocated<F>(&'static self, task: HeapBox<Task<F>>) -> JoinHandle<F::Output>
-    where
-        F: Future + 'static,
-    {
-        self.inner.scheduler.spawn_allocated::<F, HBStorage>(task)
-    }
+    // pub fn spawn_allocated<F>(&'static self, task: Box<Task<F>>) -> JoinHandle<F::Output>
```
jamesmunns, line 264: @hawkw do you have thoughts here about using the builtin `LocalScheduler` vs (re-)making a `LocalStaticScheduler` that could be allocated with the failable interface?

source/alloc/src/lib.rs
```diff
@@ -1,5 +1,420 @@
...
-pub mod containers;
-pub mod heap;
-pub mod node;
+// pub mod containers;
+// pub mod heap;
+// pub mod node;
+
+extern crate alloc;
+
+pub mod fornow {
```
jamesmunns, line 9: I plan to flatten this down in sort of the old layout, I wanted to keep the old one around for comparison until it worked. I'll remove the old allocator impl and replace it with this module soon.

source/alloc/src/heap.rs
```diff
@@ -1,505 +1,128 @@
...
     }
-}
 
-impl Drop for HeapGuard {
-    fn drop(&mut self) {
-        self.aheap.state.store(AHeap::INIT_IDLE, Ordering::SeqCst)
+    unsafe fn dealloc(&self, ptr: *mut u8, layout: core::alloc::Layout) {
+        self.allocator.dealloc(ptr, layout);
+        INHIBIT_ALLOC.store(false, Ordering::SeqCst); // TODO
+        OOM_WAITER.wake_all();
```
jamesmunns, line 126: TODO: should probably call `INHIBIT_ALLOC.swap(false, ..)` instead, and only hit the waker if it was previously true

source/alloc/src/heap.rs
```diff
@@ -1,505 +1,128 @@
...
-            let aheap_offset = cursor.align_offset(aheap_layout.align());
-            let aheap_size = aheap_layout.size();
-            used += aheap_offset;
-            used += aheap_size;
-
-            if used > size {
-                return Err(());
+pub async fn alloc(layout: Layout) -> NonNull<u8> {
+    loop {
+        unsafe {
```
hawkw, line 15: very minor nit, not important: does this entire thing actually need to be inside the unsafe block? it's really just the call to `alloc::alloc::alloc` that's unsafe...

source/alloc/src/heap.rs
```diff
@@ -1,505 +1,128 @@
...
-        }
-
-        // Nope! Stick it in the free list
-        let node_ref = NodeRef { node };
-        self.freelist.enqueue(node_ref);
-    }
+#[inline(always)]
+pub unsafe fn dealloc(ptr: *mut u8, layout: Layout) {
+    alloc::alloc::dealloc(ptr, layout)
+}
```
hawkw, line 30: nit, take it or leave it (not important): can't this just be 
```suggestion
pub use alloc::alloc::dealloc;
```

source/alloc/src/heap.rs
```diff
@@ -1,505 +1,128 @@
...
 
-    pub fn poll(&'static self) {
-        let mut hg = self.lock().unwrap();
+pub trait UlAlloc {
```
hawkw, line 32: what does `Ul` mean here?

source/alloc/src/heap.rs
```diff
@@ -1,505 +1,128 @@
...
+    const INIT: Self;
+    unsafe fn init(&self, start: NonNull<u8>, len: usize);
+    unsafe fn alloc(&self, layout: core::alloc::Layout) -> *mut u8;
+    unsafe fn dealloc(&self, ptr: *mut u8, layout: core::alloc::Layout);
+}
 
-        // Clean any pending allocs
-        hg.clean_allocs();
+pub struct Mlla {
```
hawkw, line 39: nit, sorry: can we call this something a little more...understandable? i figured out that it stands for "Mutex Linked List Alloc", but the line `impl UlAlloc for Mlla` still kind of feels like gibberish tbh...

source/alloc/src/heap.rs
```diff
@@ -1,505 +1,128 @@
...
-            }
+#[cfg(feature = "use-std")]
+impl UlAlloc for std::alloc::System {
+    const INIT: Self = std::alloc::System;
 
-            // Didn't succeed, wait until we've done some de-allocations
-            self.heap_wait.wait().await.unwrap();
-        }
+    unsafe fn init(&self, _start: NonNull<u8>, _len: usize) {
+        panic!("Don't initialize the system allocator.");
```
hawkw, line 83: would it make more sense for this to just be a nop rather than panicking, so we can have cross-platform kernel init code use it? if we're not going to call `init` in a generic context, we could also just make it be a method on the particular allocator types, rather than a trait method...

source/alloc/src/heap.rs
```diff
@@ -1,505 +1,128 @@
...
-pub unsafe fn deallocate_raw(ptr: NonNull<()>, layout: Layout) {
-    let ptr = ActiveUnsized::from_raw(ptr, layout);
-    ActiveUnsized::yeet(ptr, layout);
-}
-
-/// A guard type that provides mutually exclusive access to the allocator as
-/// long as the guard is held.
-pub struct HeapGuard {
-    aheap: &'static AHeap,
+pub struct AHeap2<U> {
```
hawkw, line 95: nitpicky, sorry: since we deleted the original `AHeap` type, can't this just be `AHeap`, rather than `AHeap2`?

source/alloc/src/heap.rs
```diff
@@ -1,505 +1,128 @@
...
-            let (start, count) = ActiveArr::<MaybeUninit<T>>::data(aa_ptr);
-            let start = start.as_ptr();
-            for i in 0..count {
-                start.add(i).write(MaybeUninit::uninit());
-            }
+unsafe impl<U: UlAlloc> GlobalAlloc for AHeap2<U> {
+    #[inline(always)]
+    unsafe fn alloc(&self, layout: core::alloc::Layout) -> *mut u8 {
+        if INHIBIT_ALLOC.load(Ordering::SeqCst) {
+            // TODO
```
hawkw, line 113: ?

source/alloc/tests/smoke.rs
hawkw, line 1: it seems like we should be able to keep some of these tests around, right?

source/kernel/src/lib.rs
```diff
@@ -331,31 +245,15 @@ impl Kernel {
...
     }
 
     pub async fn spawn<F>(&'static self, fut: F) -> JoinHandle<F::Output>
     where
         F: Future + 'static,
     {
-        let task = Task(MaitakeTask::new(fut));
-        let atask = self.heap().allocate(task).await;
-        self.spawn_allocated(atask)
+        // TODO: allocator await?
```
hawkw, line 255: we could do this by allocating the box ourselves and using `maitake`'s `spawn_allocated` with `BoxStorage`, which is provided by maitake out of the ... box. :)

source/melpomene/src/sim_drivers/emb_display.rs
```diff
@@ -268,48 +261,53 @@ impl DisplayInfo {
...
+        let fidx = self.frame_idx;
+        self.frame_idx = self.frame_idx.wrapping_add(1);
 
-            let size = (width * height) as usize;
+        match self.frames.try_push(FrameInfo { frame: fidx }) {
+            Ok(_) => {}
+            Err(_) => {
+                return Err(FrameError::NoFrameAvailable);
+            }
+        }
```
hawkw, line 272: tiny unimportant nit: could just be 
```suggestion
        self
            .frames.try_push(FrameInfo { frame: fidx })
            .map_err(|_| FrameError::NoFrameAvailable)?;
```

source/melpomene/src/sim_drivers/emb_display.rs
```diff
@@ -268,48 +261,53 @@ impl DisplayInfo {
...
-        let found = self
-            .frames
-            .iter_mut()
-            .find(|f| matches!(f, Some(FrameInfo { frame }) if *frame == frame_id));
-
-        if let Some(slot) = found {
-            *slot = None;
+        let mut found = false;
+        unsafe {
+            // This only removes items, and will not cause a realloc
```
hawkw, line 303: nit, maybe:
```suggestion
            // safety: This only removes items, and will not cause a realloc
```

source/alloc/Cargo.toml
```diff
@@ -24,3 +24,7 @@ features = ["defmt-impl"]
 [dependencies.linked_list_allocator]
 version = "0.10.1"
 default-features = false
+
+[features]
+default = []
+use-std = []
```
hawkw, line 30: nit, tioli: i feel like most of our other crates just call this feature flag "std" rather than "use-std"...

source/kernel/src/drivers/serial_mux.rs
```diff
@@ -355,33 +345,34 @@ impl IncomingMuxerTask {
...
-                    } else {
-                        warn!("Overfilled accumulator");
-                        self.idx = 0;
+                    match self.buf.try_extend_from_slice(ch) {
+                        Ok(_) => {}
+                        Err(_) => {
+                            warn!("Overfilled accumulator");
+                            self.buf.clear();
+                        }
                     }
```
hawkw, line 354: very minor, unimportant style nit: could just be 
```suggestion
                    if self.buf.try_extend_from_slice(ch).is_err() {
                         warn!("Overfilled accumulator");
                         self.buf.clear();
                    }
```

source/kernel/src/drivers/serial_mux.rs
```diff
@@ -355,33 +345,34 @@ impl IncomingMuxerTask {
...
-                        self.idx = 0;
-                        continue;
+                    match self.buf.try_extend_from_slice(ch) {
+                        Ok(_) => {}
+                        Err(_) => {
+                            warn!("Overfilled accumulator");
+                            self.buf.clear();
+                            continue;
+                        }
                     }
```
hawkw, line 373: nit, TIOLI:
```suggestion
                    if self.buf.try_extend_from_slice(ch).is_err() {
                        warn!("Overfilled accumulator");
                        self.buf.clear();
                        continue;
                    }
```

source/kernel/src/forth/mod.rs
```diff
@@ -615,7 +611,12 @@ impl BagOfHolding {
             if self.idx == 0 {
                 continue;
             }
-            if !self.inner.iter().any(|(idx, _)| *idx == self.idx) {
+            if !self
+                .inner
+                .as_slice()
```
hawkw, line 616: nit, TIOLI: it seems like this could be a `Deref` impl?

source/alloc/src/heap.rs
hawkw, line 126: a thought for the future: eventually, we probably shouldn't wake *every* task that's waiting on an allocation in `dealloc`, since the freed memory will probably not be sufficient to satisfy every outstanding allocation request. Instead, it would be nice to have the waiting node for each allocation request store the layout it desires, and then traverse the linked list until we find a waiter that can be satisfied by the freed layout, subtract its desired layout from the freed block, and (if there's still space in the layout) continue waking tasks.

but, this is substantially more complex, since it can't easily be modeled with `maitake`'s existing synchronization primitives (although, it could be a potential extension to `WaitMap` maybe?). for now, this is fine, but it's maybe worth revisiting.

hawkw (commented):
overall, this looks good and is surprisingly non-sketchy! i left a bunch of nitpicky comments. 

one major outstanding issue is that we should probably have a way to allocate a `Box<dyn Trait>`, which AFAICT is impossible with the current design...

source/alloc/src/containers.rs
```diff
@@ -1,363 +1,278 @@
...
+        // TODO: Failable way of allocating arcs?
+        Ok(Self {
+            inner: alloc::sync::Arc::new(t),
+        })
+    }
 
-    fn deref(&self) -> &Self::Target {
-        unsafe { &*Active::<T>::data(self.ptr).as_ptr() }
+    pub async fn new(t: T) -> Self {
+        // TODO: Async way of allocating Arc?
```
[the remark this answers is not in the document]
jamesmunns: Unfortunately there's no way to do that yet?

source/alloc/Cargo.toml
jamesmunns:
```
➜  mnemos git:(james/wtf-alloc) rg -i '("use-std"|use-std =)'
tools/dumbloader/Cargo.toml
10:features = ["use-std"]

tools/f3repl/Cargo.toml
12:features = ["use-std"]

source/melpomene/Cargo.toml
33:features = ["use-std"]

source/forth3/Cargo.toml
15:use-std = []

source/forth3/src/lib.rs
3:#![cfg_attr(not(any(test, doctest, feature = "use-std")), no_std)]
14:#[cfg(any(test, doctest, feature = "use-std"))]

source/forth3/src/vm/builtins/floats.rs
66:    #[cfg(feature = "use-std")]
74:    #[cfg(not(feature = "use-std"))]

source/alloc/Cargo.toml
30:use-std = []

source/alloc/src/lib.rs
1:#![cfg_attr(not(feature = "use-std"), no_std)]

source/alloc/src/heap.rs
78:#[cfg(feature = "use-std")]
```

```
$ rg -i '("std"|std =)'

source/melpomene/Cargo.toml
89:features = ["std", "async-await"]

source/trace-proto/Cargo.toml
7:std = ["tracing-serde-structured/std", "serde/std"]

source/trace-proto/src/lib.rs
1:#![cfg_attr(not(feature = "std"), no_std)]

tools/crowtty/Cargo.toml
37:features = ["std"]
```

I think this boils down to "eliza uses "std", james uses "use-std"

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -23,23 +29,30 @@ const HEAP_SIZE: usize = 384 * 1024 * 1024;
...
+#[global_allocator]
+static AHEAP: AHeap2<Mlla> = AHeap2::new();
 
 static COLLECTOR: trace::SerialCollector = trace::SerialCollector::new();
 
 /// A helper to initialize the kernel
 fn initialize_kernel() -> Result<&'static Kernel, ()> {
+    unsafe {
+        AHEAP.init(NonNull::new(AHEAP_BUF.as_ptr()).unwrap(), HEAP_SIZE);
+    }
```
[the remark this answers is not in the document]
jamesmunns: Leaving this unresolved to make sure I add docs to the kernel::new that say "the allocator must be initialized"

source/alloc/src/containers.rs
```diff
@@ -1,363 +1,278 @@
...
-impl<T> Deref for HeapBox<T> {
-    type Target = T;
+// These require the same bounds as `alloc::sync::Arc`'s `Send` and `Sync`
+// impls.
+unsafe impl<T: Send + Sync> Send for Arc<T> {}
+unsafe impl<T: Send + Sync> Sync for Arc<T> {}
+
+impl<T> Arc<T> {
+    pub fn try_new(t: T) -> Result<Self, T> {
+        // TODO: Failable way of allocating arcs?
```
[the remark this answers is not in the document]
jamesmunns: Not today :(

source/alloc/src/containers.rs
```diff
@@ -1,363 +1,278 @@
...
-    }
-}
-
-impl<T> Deref for HeapArc<T> {
-    type Target = T;
-
-    #[inline(always)]
-    fn deref(&self) -> &Self::Target {
+    pub async fn new(t: T) -> Self {
+        let ptr: *mut T = alloc(Layout::new::<T>()).await.cast().as_ptr();
```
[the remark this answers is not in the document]
jamesmunns: You uh, could do `alloc::boxed::Box::new(/*whatever*/)` then `into_raw` it then `mnemos_alloc::containers::Box::from_raw` that.

source/alloc/src/containers.rs
```diff
@@ -1,363 +1,278 @@
...
 }
 
-impl<T> fmt::Pointer for HeapArc<T> {
-    #[inline]
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        fmt::Pointer::fmt(&self.ptr, f)
-    }
+pub struct ArrayBuf<T> {
+    ptr: NonNull<UnsafeCell<MaybeUninit<T>>>,
+    _pd: PhantomData<*const T>,
```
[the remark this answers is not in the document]
jamesmunns: Ah, probably nothing. Will remove.

source/alloc/src/containers.rs
```diff
@@ -1,363 +1,278 @@
...
-    // }
 
-    /// Leak the contents of this box, never to be recovered (probably)
-    pub fn leak(self) -> (NonNull<T>, usize) {
-        unsafe {
-            let (nn_ptr, count) = ActiveArr::<T>::data(self.ptr);
-            forget(self);
-            (nn_ptr, count)
-        }
+    pub fn ptrlen(&self) -> (NonNull<UnsafeCell<MaybeUninit<T>>>, usize) {
```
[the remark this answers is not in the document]
jamesmunns: It's not really `into`, it doesn't represent a "leak" or whatever (maybe it should?), more `as_raw_parts`?

source/alloc/tests/smoke.rs
jamesmunns: I'm not sure how much we need to test the underlying allocator. The original code was testing the allocator implementation, which doesn't really exist anymore.

source/kernel/src/forth/mod.rs
jamesmunns: I think there was a reason I didn't do that, maybe to leave us open to deref/derefmut something else? I'll think about while I'm doc'ing.

source/kernel/src/lib.rs
jamesmunns: I tried doing that, and it caused a lot of problems, but that was very mid-switchover, maybe it'd be nicer now.

source/kernel/src/lib.rs
jamesmunns: I think this spawns tasks, not the scheduler. So if we spawned a new task 6 hours after startup, this *could* panic there.

source/alloc/README.md
```diff
@@ -2,88 +2,6 @@
...
-
-## That sounds terribly unsafe!
-
-Oh it is! Egregiously! See the reviews below:
-
-> "this code is like a fractal footgun forest"
->
-> - [@dirbaio]
-
-[@dirbaio]: https://twitter.com/Dirbaio/
```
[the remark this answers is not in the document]
jamesmunns: Yeah, that's fair - some of it can be brought back. Most of it was describing the tricky type shenanigans I was doing that I no longer am doing.

source/alloc/Cargo.toml
hawkw: lol, welp. maybe we should make this consistent...

source/alloc/src/heap.rs
```diff
@@ -1,505 +1,213 @@
...
+///     * If we are NOT inhibited, but are now out of memory (the underlying
+///       allocator returned a nullptr), we inhibit further allocations until
+///       the next deallocation occurs
+/// * On **dealloc**:
+///     * The "inhibit allocations" flag is cleared
+///     * If any tasks are waiting on the "OOM" queue, they are ALL awoken if
+///       the inhibit flag was previously set
+///
+/// These two details are intended to allow the "async allocation aware" types
+/// defined in [crate::containers] to yield if allocation is not currently possible.
```
hawkw, line 34: love this!

hawkw (approved):
docs are lovely!

PR #99, opened 2023-06-25
feat(melpo): trace improvement

hawkw:
This commit demotes the traces emitted by Melpomene on every iteration of the kernel run loop from DEBUG to TRACE, as these occur *extremely* frequently and spam the console when trying to get debugging output from other parts of the kernel. In addition, I've added new tracing messages when a kernel server is started in Melpomene so that we can easily trace the boot process.

### DIFF
--- a/source/melpomene/src/main.rs
+++ b/source/melpomene/src/main.rs
@@ -96,39 +96,58 @@ async fn kernel_entry(opts: MelpomeneOptions) {
             //
             // Create the buffer, and spawn the worker task, giving it one of the
             // queue handles
+            tracing::debug!("initializing simulated UART ({})", opts.serial_addr);
             TcpSerial::register(k, opts.serial_addr, 4096, 4096, irq)
                 .await
                 .unwrap();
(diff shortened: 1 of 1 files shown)


source/melpomene/src/main.rs
```diff
@@ -96,39 +96,58 @@ async fn kernel_entry(opts: MelpomeneOptions) {
...
-        // * Framed messages up to 512 bytes max each
-        SerialMuxServer::register(k, 16, 512).await.unwrap();
-    })
+    k.initialize(
+        async {
+            // * Up to 16 virtual ports max
+            // * Framed messages up to 512 bytes max each
+            let ports = 16;
+            let frame_size = 512;
+            tracing::debug!(ports, frame_size, "initializing SerialMuxServer...");
```
jamesmunns, line 115: nit: maybe we do this inside the servers or as `instrument` macros instead?

jamesmunns (approved):
One nit, TIOLI

PR #100, opened 2023-06-25
feat(crowtty): make trace format match melpomene

hawkw:
This commit changes the `crowtty` trace format slightly so that it's a closer match to the `tracing_subscriber::fmt` format used by Melpomene. This, admittedly, does not actually matter at all, but I wanted to do it.


jamesmunns (approved):
> This, admittedly, does not actually matter at all, but I wanted to do it.

I will never say no to QoL improvements (real or perceived)

PR #103, opened 2023-06-25
feat(d1): implement I2C driver

### LINKED ISSUE #101: kernel: service definition for I2C
The kernel should include generic `I2cService` and `I2cClient` definitions, similarly to how we have generic definitions for `SimpleSerialService` but no actual driver task definition.

### LINKED ISSUE #102: platform(d1): implement I2C service with a TWI driver
The D1 driver crate should have an I2C server that implements the kernel I2C service definition (see #101) using the D1's TWIs.

hawkw:
## Motivation

The Beepy keyboard is driven by firmware (called `i2c_puppet`) running
on an RP2040 MCU, which can be communicated with over the I<sup>2</sup>C
bus. In order to use the keyboard, we therefore must implement a driver
for the Allwinner D1's I<sup>2</sup>C controller.

Closes #101
Closes #102

## Solution

This branch adds a definition for an I<sup>2</sup>C driver service to
`kernel::services`, and an implementation of the I<sup>2</sup>C driver
service for the Allwinner D1's TWI hardware.

### I<sup>2</sup>C Service

The cross-platform kernel cannot implement the server for the
I<sup>2</sup>C driver, because the driver implementation is
platform-specific; but we can provide an abstraction for interfacing
with drivers for a platform's I<sup>2</sup>C hardware. The
I<sup>2</sup>C service provides an `I2cClient` type, which is used to
initiate bus operations an I<sup>2</sup>C target device at a provided
address. Once a bus transaction is initiated, a `Transaction` type is
performed, which can be used to perform multiple read and write
operations with the target device. When the `Transaction` is completed,
a `STOP` condition is sent on the bus; until then, completing a read or
write will instead send a repeated-`START`. This interface permits the
common I<sup>2</sup>C pattern of writing a register address to read
from, and then reading the contents of that register from the target
within the same transaction.

In addition, the transaction interface can be used to implement the
`embedded_hal_async` crate's `I2c` trait for the `I2cClient` type. This
provides a higher level interface for I<sup>2</sup>C operations, and
allows our I<sup>2</sup>C driver to be used with third-party libraries
implemented against the `embedded_hal_async` interface.

Because of mnemOS' message-passing design, the `I2cService` operates
with owned buffers, rather than borrowed buffers, so a `FixedVec<u8>` is
used as the buffer type for both read and write operations. This means
that we must allocate when performing I<sup>2</sup>C operations. To
reduce the amount of allocation necessary, all `Transaction` methods
return the buffer that was passed in, allowing the buffer to be reused
for multiple operations. To facilitate this, the `Transaction::read` and
`Transaction::write` methods also take a `len` parameter indicating the
actual number of bytes to write from the buffer/read into the buffer,
rather than always writing the entire buffer contents or filling the
entire buffer with bytes. This way, we can size the buffer to the
largest buffer required for a sequence of operations, but perform
smaller reads and writes using the same `FixedVec<u8>`, avoiding
reallocations. The implementation of
`embedded_hal_async::i2c::I2c::transaction` will allocate a single
buffer large enough for the largest operation in the transaction, and
reuse that buffer for every operation.

### TWI driver

This branch adds an implementation of a driver for controlling the
I<sup>2</sup>C hardware on the D1, which the D1 manual calls a TWI
(Two-Wire Interface), likely due to I<sup>2</sup>C being a trademark of
Phillips Semiconductor. The TWI hardware can be used in one of two
modes: "TWI engine" mode, where individual bytes are written/read from
the I<sup>2</sup>C bus in an interrupt handler, and "TWI driver" mode,
where the TWI hardware can operate at the level of I<sup>2</sup>C
register read/writes using a DMA buffer. This branch only implements a
driver for the TWI engine mode, since it can model all forms of
I<sup>2</sup>C operations. In the future, we will likely want to
opportunistically use the offload capabilities of the TWI driver when
the I<sup>2</sup>C transaction has the correct shape for offloading, but
this branch just implements the simpler TWI engine mode.

The TWI hardware is a bit difficult to use correctly, so implementing
this was a bit of a struggle. In particular, it turns out that the
generation of I<sup>2</sup>C clock pulses occurs when the `TWI_CNTR`
register, which controls the TWI, is written to.

The driver works by sharing state between a driver task and an ISR,
since the TWI engine mode is interrupt driven. The shared state is
"locked" by disabling TWI interrupts temporarily while the driver task
is writing to the shared state, and resuming interrupts when a write to
the shared state completes. In theory, this driver is also safe for use
on multi-core hardware, since a single daemon task is responsible for
all the non-ISR writes to this shared state, but this doesn't actually
matter, because the D1 is an inherently single-core CPU.

The D1 has four separate TWI controllers, `TWI0`, `TWI1`, `TWI2`, and
`TWI3`. The pin mapping for the 40-pin Pi header `I2C0` pins differs
between the MangoPi MQ Pro and the Lichee RV Dock. On the MQ Pro, `TWI0`
is used, with `SCL` on pin `PG12` and `SDA` on pin `PG13`. On the Lichee
RV Dock, `TWI2` is used instead, with `SCL` on pin `PB0` and `SDA` on
pin `PB1`. Therefore, the respective pin mappings are set up by the
individual board support targets, rather than in `mnemos_d1_core`.

### Validation

I've validated the implementation by testing with the Beepy's
`i2c_puppet` device. Now, when running on the Beepy, we spawn a task
that performs a simple read operation to get `i2c_puppet`'s firmware
version, and then sends commands to `i2c_puppet` to turn the Beepy's RGB
LED green. This works :)

image
image

## Future Work

Some things this branch *doesn't* do, but that we will want to do in the
future, include:

- [ ] Implement support for the "TWI driver" DMA offload mode, as
  discussed above
- [ ] Implement 10-bit I<sup>2</sup>C addresses. The current
  implementation only supports 7-bit address mode. (#143)
- [ ] Potentially, add a driver for MnemOS operating as the
  I<sup>2</sup>C bus target device rather than as the bus controller?
  I'm not sure if this is actually something we would ever need,
  though...
- [ ] Implement a proper driver for `i2c_puppet`, so that we can use the
  keyboard.
- [ ] A potential optimization involving an `I2cClient` reusing the same
  allocated buffer across multiple `embedded_hal_async` `I2c::transaction`s
  could be worth implementing.


source/kernel/src/registry/mod.rs
```diff
@@ -27,13 +27,15 @@ pub mod known_uuids {
         pub const SERIAL_MUX: Uuid = uuid!("54c983fa-736f-4223-b90d-c4360a308647");
         pub const SIMPLE_SERIAL_PORT: Uuid = uuid!("f06aac01-2773-4266-8681-583ffe756554");
         pub const EMB_DISPLAY: Uuid = uuid!("b54db574-3eb7-4c89-8bfb-1a20890be68e");
+        pub const I2C: Uuid = uuid!("011ebd3e-1b14-4bfd-b581-6138239b82f3");
```
jamesmunns, line 30: Doesn't need to be fixed in this PR, but we *probably* want to think about including date/version values here. Eventually we should care about breaking interface changes.

hawkw: yup, good idea!

platforms/allwinner-d1/boards/lichee-rv/src/twi.rs
```diff
@@ -0,0 +1,571 @@
...
+                byte_counter: len as usize,
+                link: None,
+                wait_clock_cycles: 0,
+                bmode: BModeSel::Normal,
+                dest_width: DataWidth::Bit8,
+                dest_addr_mode: AddressMode::LinearMode,
+                dest_block_size: BlockSize::Byte1,
+                dest_drq_type: DestDrqType::Dram,
+                src_data_width: DataWidth::Bit8,
+                src_addr_mode: AddressMode::LinearMode,
```
jamesmunns, line 269: This should probably be IoMode

platforms/allwinner-d1/boards/src/beepy.rs
```diff
@@ -0,0 +1,102 @@
...
+        })
+        .unwrap();
+}
+
+/// A rudimentary driver for Beepy's [i2c_puppet].
+///
+/// [i2c_puppet]: https://beepy.sqfmi.com/docs/firmware/keyboard
+// TODO(eliza): turn this into a service!
+#[trace::instrument(skip(k))]
+pub async fn i2c_puppet(k: &'static Kernel) -> Result<(), I2cError> {
```
jamesmunns, line 30: Have I mentioned how cool this is?

hawkw: :D

platforms/allwinner-d1/core/src/drivers/twi.rs
```diff
@@ -0,0 +1,809 @@
...
+        Addr, I2cService, Transaction,
+    },
+    trace, Kernel,
+};
+
+/// A TWI mapped to the Raspberry Pi header's I<sup>2</sup>C0 pins.
+pub struct I2c0 {
+    isr: &'static IsrData,
+    twi: &'static twi::RegisterBlock,
+    /// Which TWI does this TWI Engine use?
```
jamesmunns, line 73: At first I thought this was a todo comment, but it's actually saying "this field owns whatever Interrupt and handler is used for the instance".

But if you can have multiple instances, why is it called I2C0?

platforms/allwinner-d1/core/src/drivers/twi.rs
```diff
@@ -0,0 +1,809 @@
...
+        // disable TWI interrupts while holding the guard.
+        twi.twi_cntr.modify(|_r, w| w.int_en().low());
+        let data = unsafe { &mut *(self.data.get()) };
+        TwiDataGuard { data, twi }
+    }
+}
+
+impl Drop for TwiDataGuard<'_> {
+    fn drop(&mut self) {
+        self.twi.twi_cntr.modify(|_r, w| w.int_en().high());
```
jamesmunns, line 442: Should we check if the int *wasn't* enabled before locking? not sure if this matters, but locking and unlocking could indirectly enable the interrupt even if you weren't expecting it.

platforms/allwinner-d1/core/src/drivers/twi.rs
```diff
@@ -0,0 +1,809 @@
...
+}
+
+impl DerefMut for TwiDataGuard<'_> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.data
+    }
+}
+
+impl TwiData {
+    fn advance_isr<const NUM: u8>(&mut self, twi: &twi::RegisterBlock) {
```
jamesmunns, line 491: Just because this function is sorta big and I don't understand what NUM is used for, any reason to make this generic over NUM instead of taking it as an arg?

I dunno if const generics monomorphize bloat like other functions, but if we need the generic, we might want to make an inner function that isn't generic to keep code size down.

platforms/allwinner-d1/core/src/lib.rs
```diff
@@ -72,7 +75,7 @@ impl D1 {
                 .unwrap()
         };
 
-        let [ch0, ..] = dmac.channels;
+        let [ch0, _, ..] = dmac.channels;
```
jamesmunns, line 78: Maybe put this back?

platforms/allwinner-d1/core/src/lib.rs
```diff
@@ -278,9 +297,12 @@ impl D1 {
...
     /// DMAC ISR handler
     ///
-    /// At the moment, we only service the Channel 0 interrupt,
-    /// which indicates that the serial transmission is complete.
+    /// At the moment, we service the interrupts on the following channels:
+    /// * Channel 0: UART0 TX
+    /// * Channel 1: SPI1 TX
+    /// * Channel 2: TWI0 driver TX
     fn handle_dmac() {
+        let _isr = kernel::isr::Isr::enter();
```
jamesmunns, line 305: Might want a comment to explain what this is doing (swapping tracing buffers)

source/alloc/src/heap.rs
```diff
@@ -85,6 +85,8 @@ static OOM_WAITER: WaitQueue = WaitQueue::new();
 /// that *could* potentially succeed
 static INHIBIT_ALLOC: AtomicBool = AtomicBool::new(false);
 
+pub static OOM_PANIC: AtomicBool = AtomicBool::new(false);
```
jamesmunns, line 88: Do we ever set this true?

platforms/allwinner-d1/core/src/drivers/twi.rs
hawkw: theoretically yeah, we probably should — its not actually an issue here because the guard type is only used in one place, but it would be more correct to do that.

platforms/allwinner-d1/core/src/drivers/twi.rs
hawkw: i did the silly const generic thing because i thought there might be a performance benefit from not passing an extra arg that's only used for logging. but, code size probably has a bigger impact on performance than passing one integer in a register, so you're right, this is probably dumb!

platforms/allwinner-d1/core/src/lib.rs
hawkw: whoops

platforms/allwinner-d1/core/src/lib.rs
hawkw: yeah, i actually think this line can be removed, since the PR that added the "in_isr" tracking just put it in the `MachineExternal` interrupt function. this is just a rebase artifact.

source/kernel/src/services/i2c.rs
```diff
@@ -0,0 +1,685 @@
...
+            match op {
+                i2c::Operation::Read(dest) => {
+                    let len = dest.len();
+                    let read = txn.read(buf, len, end).await?;
+                    dest.copy_from_slice(read.as_slice());
+                    buf = read;
+                }
+                i2c::Operation::Write(src) => {
+                    let len = src.len();
+                    buf.try_extend_from_slice(src)
```
jamesmunns, line 381: Could we check the capacity ahead of time instead of panicking?

source/kernel/src/services/i2c.rs
```diff
@@ -0,0 +1,685 @@
...
+        let cap = buf.capacity() - buf.len();
+        if cap < len {
+            return Err(self.buf_too_small(OpKind::Read, len, cap));
+        }
+        self.xfer(buf, len, end, OpKind::Read).await
+    }
+
+    /// Write `len` bytes `buf` to the I²C.
+    ///
+    /// Note that, rather than always writing all the bytes in the buffer, this
```
jamesmunns, line 465: I'm less certain we need this extra len for write, is there a case you wouldn't want to "send all" for write?

For read it makes sense: there is no len, only capacity. For write, can't we just use len?

source/alloc/src/heap.rs
hawkw: agh, this was a temporary debugging thing, i meant to remove it --- good catch.

source/kernel/src/services/i2c.rs
hawkw: hmm, that's a good point, we probably could remove this. good call!

jamesmunns (approved):
Overall looks awesome, left a bunch of nits, feel free to take them or leave them :)

source/kernel/src/services/i2c.rs
hawkw: we allocate the buffer with enough capacity to fit the biggest `i2c::Operation`'s buffer, see https://github.com/tosc-rs/mnemos/blob/10185233cda73faea03716e609182c84aa0c422e/source/kernel/src/services/i2c.rs#L355-L366

if the `try_extend_from_slice` call fails, it means one of the following:
- `buf.clear()` decided not to clear the buffer for some reason
- the allocator decided to give us a buffer smaller than the one we asked for, just to fuck with us
- one of the `i2c::Operation`s was resized while we had an `&mut` reference to it (someone is doing something evil)

so, all of those suggest something Really Fucked Up is happening and i think it's reasonable to panic.

platforms/allwinner-d1/core/src/drivers/twi.rs
hawkw: You can't have multiple instances --- constructing an `I2c0` consumes the `TWI`
from Peripherals, so you can't call it multiple times without passing in
ownership of that peripheral. Which TWI corresponds to the I2C0 pins on the Pi
header depends on the board, which is why we need to have this, to tell the
board-agnostic initialization code which interrupt to unmask.

platforms/allwinner-d1/core/src/drivers/twi.rs
hawkw: really, the `MutexGuard`-like behavior is not actually necessary any longer. at one point, i was actually locking and dropping the guard in several places. now, the only time we actually unmask the interrupt is inside of `wait_for_irq`, which guarantees that the driver task is not writing to the shared state just because it's yielded to wait for the IRQ. so, we could refactor this to remove the guard-like behavior...

PR #105, opened 2023-06-27
Extract more melpo components (and also rename a bunch of stuff)

jamesmunns:
This PR pulls some boilerplate functionality out of melpo, to make it easier to reuse
in other places, such as on hardware projects.

This also makes some somewhat significant renames, moving "servers" and "shells" to "daemons", and renaming "drivers" to "services".


source/kernel/src/servers.rs
```diff
@@ -0,0 +1,100 @@
+//! Servers
```
jamesmunns, line 1: I couldn't come up with a better name here, maybe "non-registry servers/services"?

Like, these are more "do a thing" background tasks than "provide a capability" tasks, like services.

hawkw: "daemons", maybe?

platforms/allwinner-d1/boards/lichee-rv/src/main.rs
```diff
@@ -196,58 +199,41 @@ fn main() -> ! {
...
 
-        loop {
-            hdl.send(b"Hello, world!\r\n").await;
-            k.sleep(Duration::from_secs(1)).await;
-        }
-    })
-    .unwrap();
+    // Spawn a hello port
+    let hello_settings = SermuxHelloSettings::default();
+    k.initialize(sermux_hello(k, hello_settings)).unwrap();
```
hawkw, line 236: nit, not a big deal: maybe these should go in a `services::sermux` module, so we can reference them as `sermux::loopback` and sermux::hello`?

source/kernel/src/servers.rs
```diff
@@ -0,0 +1,100 @@
...
+    tracing, Kernel,
+};
+
+//
+// Sermux Loopback
+//
+
+/// Loopback Settings
+#[derive(Debug, Clone)]
+pub struct SermuxLoopbackSettings {
```
hawkw, line 18: tiny nit, not actually important but i thought i'd mention it given that you've put `#[non_exhaustive]` on the `WellKnown` enum: maybe this should have a private `()` field, so that it can only be constructed as
```rust
SermuxLoopbackSettings {
    port: 0,
    buffer_size: 128,
    ...Default::default(),
}
```
so that if we add more settings in the future, existing code doesn't break, but it's explicit that it's getting the defaults for any new settings?

source/kernel/src/servers.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+//
+// Sermux Hello
+//
+
+/// Hello Server Settings
+#[derive(Debug, Clone)]
+pub struct SermuxHelloSettings {
+    /// Port number. Defaults to [WellKnown::Loopback]
```
hawkw, line 60:
```suggestion
    /// Port number. Defaults to [WellKnown::HelloWorld]
```

source/kernel/src/servers.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+//
+// Sermux Hello
+//
+
+/// Hello Server Settings
+#[derive(Debug, Clone)]
+pub struct SermuxHelloSettings {
```
hawkw, line 59: same comment as above; perhaps this ought to have a "no construct without default" field?

source/melpomene/src/main.rs
```diff
@@ -112,76 +113,29 @@ async fn kernel_entry(opts: MelpomeneOptions) {
...
-            .await;
-        }
-        .instrument(tracing::info_span!("Graphics Console")),
-    )
+    k.initialize(graphical_shell_mono(
+        k,
+        DISPLAY_WIDTH_PX,
+        DISPLAY_HEIGHT_PX,
+        WellKnown::PsuedoKeyboard as u16,
+        1024, // capacity
```
hawkw, line 137: nit, take it or leave it: maybe we ought to add a `GraphicalShellSettings` type, similarly to the sermux stuff?

hawkw (approved):
this looks great! i left a few comments

source/melpomene/src/main.rs
jamesmunns: Maybe, I dunno if there's a "right" default for things like display size. Right now it's all 400x240, but that's an artifact of "only supporting one display" thing.

PR #106, opened 2023-06-27
Make spawnulator a service

jamesmunns:
This PR moves the spawnulator into a kernel service, allowing for potentially
"non-hierarchical" forth instances (e.g. a tid0 gui and tid1 sermux instance) to
use the same spawnulator.


jamesmunns:
There is a *slight* footgun, in that it is possible to try to spawn forth tasks without
having spawned the spawnulator, which I did run into during testing.

Open to suggestions on how to fix this, including maybe lazily/racily spawning the spawnulator instead of doing the normal "wait until the server appears" behavior.

source/kernel/src/forth/mod.rs
```diff
@@ -385,7 +388,7 @@ async fn spawn_forth_task(forth: &mut forth3::Forth<MnemosContext>) -> Result<()
         );
         forth3::Error::InternalError
     })?;
-    let host_ctxt = MnemosContext::new(kernel, params, forth.host_ctxt.spawnulator.clone()).await;
+    let host_ctxt = MnemosContext::new(kernel, params).await;
     let child_id = host_ctxt.id;
```
jamesmunns, line 392: I'm not sure where we actually increment this (I didn't look super hard - will do tomorrow), but we might want to use a static atomic for the count (see pr notes about "non hierarchical spawns"), or go ahead and use the maitake task ID instead.

source/kernel/src/drivers/mod.rs
```diff
@@ -17,5 +17,6 @@
 //!       the [serial_mux] module.
 
 pub mod emb_display;
+pub mod forth_spawnulator;
```
hawkw, line 20: nit: does this really belong in `drivers`? maybe we need a `mod services` or something?

source/kernel/src/drivers/forth_spawnulator.rs
```diff
@@ -0,0 +1,175 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+// Service Definition
+////////////////////////////////////////////////////////////////////////////////
+pub struct Spawnulator;
+
+impl RegisteredDriver for Spawnulator {
+    type Request = Request;
+    type Response = Response;
+    type Error = Error;
```
hawkw, line 65: afaict we never return an error, maybe this should be `core::convert::Infallible`?

hawkw (approved):
always be spawnulating

source/kernel/src/drivers/forth_spawnulator.rs
jamesmunns: I didn't think about that, but that is a very good idea!

source/kernel/src/drivers/mod.rs
jamesmunns: Yeah... I dunno what the naming/hierarchy for this should be. It probably should be services.

jamesmunns:
> There is a slight footgun, in that it is possible to try to spawn forth tasks without
having spawned the spawnulator, which I did run into during testing.

I ran into this again while developing #110, which means I should definitely fix it.

source/kernel/src/drivers/mod.rs
hawkw: not a big deal, just thought i'd mention it

hawkw:
> > There is a slight footgun, in that it is possible to try to spawn forth tasks without
> 
> having spawned the spawnulator, which I did run into during testing.
> 
> 
> 
> I ran into this again while developing #110, which means I should definitely fix it.

maybe we could have an atomic flag that gets set when the spawnulator is spawnulated? and check that every time we spawn a forth task. minor overhead but it's probably fine.

PR #112, opened 2023-06-29
feat(allwinner-d1): add bin targets for MQ Pro and Lichee RV

hawkw:
Currently, the MnemOS platform impl for the Allwinner D1 SoC can run on
both the MangoPi MQ Pro *or* Lichee RV boards. These boards are very
similar and most of the platform code can be shared, but some pin
mappings (e.g. LEDs, I2C) differ. It would be nice to be able to build
MnemOS for both boards, rather than having to manually change pin
mappings depending on the target you're building for.

This branch splits out most of the non-board-specific code from
`allwinner-d1/boards/lichee-rv` and moves it into the
`allwinner-d1/core` crate (formerly `allwinner-d1/drivers`). The
`boards/lichee-rv` crate is removed, and `boards/` is turned into a
crate which contains separate Cargo `[[bin]]` targets for the Lichee RV
and MQ Pro. All these bin targets do is set up the LED pins and call
into the `d1-core` crate.

In the future, when we need to add support for other hardware features
where pin mappings differ based on the SBC, such as I2C, we can add a
generic driver in `d1-core` and configure pin mappings in the
board-specific bin target

I've also done a quick documentation update pass, let me know what you
think!


hawkw:
i think i'm actually going to change this from using separate crates to use separate bin targets in the same crate

platforms/allwinner-d1/core/src/drivers/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub mod sharp_display;
```
jamesmunns, line 1: Someday this shouldn't live here, but today is not that day.

jamesmunns (approved):
Overall this looks good! I think some day we will also want to rename `boards/`, but it's totally fine for now.

PR #113, opened 2023-06-30
Do some stuff with just

jamesmunns:
I noticed some of the `just` tasks were broken due to d1-core changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,15 +31,9 @@ env:
   RUST_BACKTRACE: short
 
 jobs:
-  # run `cargo check` with the host target triple.
-  #
-  # NOTE: the `lichee-rv` crate is excluded as it is a no-std package, and when building
-  # the entire workspace, this setting isn't respected, leading to compile errors about
-  # duplicate panic-handler definitions.
-  #
-  # See the `check-allwinner-d1` job below that builds that project specifically.
-  check-host:
-    name: cargo check (host)
+  # run `just check`
+  check:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -50,42 +44,15 @@ jobs:
...
+    - uses: extractions/setup-just@v1
+    - name: run just check
       run: |
-        cargo check \
-          --workspace \
-          --all-features \
-          --quiet \
-          --message-format=json |
-          cargo-action-fmt
+        just check
```
hawkw, line 50: nit: this could just be
```yaml
- run: just check
```
now, the `|` is only needed when the string is multiline, and if we exclude the `name`, the name of the step is just the `run:` command.

.github/workflows/ci.yml
```diff
@@ -99,17 +66,16 @@ jobs:
...
       uses: taiki-e/install-action@nextest
     - uses: actions/checkout@v2
-    - name: cargo nextest run
+    - uses: extractions/setup-just@v1
+    - name: just test
       run: |
-        cargo nextest run \
-          --workspace \
-          --all-features
+        just test
```
hawkw, line 72: similarly we could replace this whole thing with `- run: just test`

hawkw (approved):
whoops, my bad, thanks for fixing this!

.github/workflows/ci.yml
jamesmunns: yaml scares me

PR #110, opened 2023-06-28
WIP: display on hardware

jamesmunns:
Very inefficient, needs a rebase (it relies on #106), but it works :)

PR #114, opened 2023-06-30
feat(d1): Add sharp display driver

jamesmunns:
This restores/extracts the experimental display driver from #110 as a replacement to the "blinky pattern" that currently exists. This driver is largely based on the `emb_display` driver that exists in melpo.

The driver is still not *great*, but we should probably rethink the `emb_display` in melpo first, and update this driver to be consistent with that.


platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -1,73 +1,393 @@
...
+
+/// Implements the [`EmbDisplayService`] service interface
+pub struct SharpDisplay;
+
+impl SharpDisplay {
+    /// Register the driver instance
+    ///
+    /// Registration will also start the simulated display, meaning that the display
+    /// window will appear.
+    pub async fn register(kernel: &'static Kernel, max_frames: usize) -> Result<(), FrameError> {
```
hawkw, line 53: eventually i imagine we'll want to construct the driver with a config type that configures the display's dimensions?

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -1,73 +1,393 @@
...
+        let frames = FixedVec::new(max_frames).await;
+        let mut linebuf = FixedVec::new((52 * 240) + 2).await;
+        for _ in 0..(52 * 240) + 2 {
```
hawkw, line 56: nit: should we use the constant here
```suggestion
        let mut linebuf = FixedVec::new((52 * HEIGHT) + 2).await;
        for _ in 0..(52 * HEIGHT) + 2 {
```

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -1,73 +1,393 @@
...
+    }
+}
+
+impl Dimensions for FullFrame {
+    fn bounding_box(&self) -> embedded_graphics::primitives::Rectangle {
+        Rectangle::new(
+            Point { x: 0, y: 0 },
+            Size {
+                width: 400,
+                height: 240,
```
hawkw, line 136: nit: should we use the constants here?
```suggestion
                width: WIDTH,
                height: HEIGHT,
```
(we might have to cast from `usize` to whatever these fields are, i think they might be `u32` but can't remember OTTOMH?)

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -1,73 +1,393 @@
...
+    buf: FixedVec<u8>,
+    spim: SpiSenderClient,
+    ctxt: Arc<Mutex<Context>>,
+}
+
+impl Draw {
+    async fn run(mut self) {
+        loop {
+            let c = self.ctxt.lock().await;
+            // render into
```
hawkw, line 202: is this an unfinished comment?

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -1,73 +1,393 @@
...
+            let vc = 0x01 | if c.vcom { 0x02 } else { 0x00 };
+            self.buf.as_slice_mut()[0] = 0x01 | vc;
```
hawkw, line 204: wait, this gets or'd with `0x01` twice, is that supposed to happen?

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
hawkw, line 56: nit, also, there are a bunch of "52"s in this file that seems like they could be const'd?

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -1,73 +1,393 @@
...
+/// been set, as a power optimization or latency improvement.
+struct Draw {
+    kernel: &'static Kernel,
+    buf: FixedVec<u8>,
+    spim: SpiSenderClient,
+    ctxt: Arc<Mutex<Context>>,
+}
+
+impl Draw {
+    async fn run(mut self) {
```
hawkw, line 199: nit, take it or leave it: maybe worth giving this a tracing span?

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -1,73 +1,393 @@
...
+    kernel: &'static Kernel,
+    ctxt: Arc<Mutex<Context>>,
+}
+
+impl VCom {
+    pub async fn run(self) {
+        loop {
+            self.kernel.sleep(Duration::from_secs(1)).await;
+            let mut c = self.ctxt.lock().await;
+            c.vcom = !c.vcom;
```
hawkw, line 179: in the future it seems like the vcom bit could maybe be an `AtomicBool`?

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -1,73 +1,393 @@
...
+/// framebuffer.
+struct CommanderTask {
+    cmd: KConsumer<Message<EmbDisplayService>>,
+    display_info: DisplayInfo,
+    ctxt: Arc<Mutex<Context>>,
+}
+
+impl CommanderTask {
+    /// The entrypoint for the driver execution
+    async fn run(mut self) {
```
hawkw, line 236: nit, take it or leave it: maybe worth giving this a tracing span?

hawkw (approved):
you love to see it! :D 

i commented on a few minor nits, feel free to ignore me.

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
jamesmunns: Not really, the sharp display only comes in two sizes, and I've never seen someone use the other one (that isn't 400x240)

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
jamesmunns: Could be, though see the note above about supporting a gpio in the future

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
jamesmunns: that's a good call

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
jamesmunns: Ah, this looks like a half finished thought. I meant to move it all to 203, and then didn't remove the command bit from 204. Good catch (it would do nothing)

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
jamesmunns: also a good call

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
jamesmunns: lol whoops

jamesmunns:
will fix the nits (unless otherwise stated) and merge tomorrow 😴

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -1,73 +1,393 @@
-use crate::drivers::spim::SpiSenderClient;
+//! Sharp display driver
+//!
+//! This is an early attempt at a "frame buffer" style display driver. It provides a
+//! [emb_display service][kernel::services::emb_display] server, and uses the
```
jamesmunns, line 4: note to self, finish this line too

PR #116, opened 2023-07-01
feat(allwinner-d1): panic handling improvements

hawkw:
This branch contains two commits:

* feat(allwinner-d1): disable IRQs in panic

  This commit changes the panic handler for the D1 to disable
  interrupts, and adds a flag for tracking double panics. If we panic
  inside the panic handler, we now just die immediately instead of
  trying to print something.

* feat(crowtty): try to decode loose utf8

  This commit changes crowtty to attempt to decode any output that isn't
  in a SerMux frame as UTF-8 text. This lets us get panic output without
  having to do fancy COBS encoding in the panic handler. It also lets us
  see other stuff, like `xfel`'s bootloader output.
  
  For example: 
  image

### DIFF
--- a/platforms/allwinner-d1/core/src/lib.rs
+++ b/platforms/allwinner-d1/core/src/lib.rs
@@ -281,12 +281,25 @@ impl D1 {
         // Ugly but works
         let mut uart: Uart = unsafe { core::mem::transmute(()) };
 
-        write!(&mut uart, "\r\n").ok();
-        write!(&mut uart, "{}\r\n", info).ok();
+        // end any existing SerMux frame on the UART
+        uart.write(&[0]);
+
+        // write out the panic message in plaintext
+        write!(&mut uart, "\r\n{info}\r\n").ok();
+        // end the SerMux frame so crowtty can decode the panic message as utf8
+        uart.write(&[0]);
+
+        write!(
+            &mut uart,
+            "you've met with a terrible fate, haven't you?\r\n"
+        )
+        .ok();
         uart.write(&[0]);
 
         die();
 
+        /// to sleep, perchance to dream; aye, there's the rub,
+        /// for in that sleep of death, what dreams may come?
         fn die() -> ! {
             loop {
                 // wait for an interrupt to pause the CPU. since we just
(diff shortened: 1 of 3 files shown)


platforms/allwinner-d1/core/src/lib.rs
```diff
@@ -260,14 +265,32 @@ impl D1 {
...
         write!(&mut uart, "{}\r\n", info).ok();
+        uart.write(&[0]);
 
-        loop {
-            core::sync::atomic::fence(Ordering::SeqCst);
+        die();
+
+        fn die() -> ! {
+            loop {
+                core::sync::atomic::fence(Ordering::SeqCst);
```
jamesmunns, line 292: Not a huge deal, but you could wfi loop here instead to not heat the CPU (it'll probably just never wake)

hawkw: oh good call

platforms/allwinner-d1/core/src/lib.rs
```diff
@@ -260,14 +265,32 @@ impl D1 {
...
+        if PANICKING.swap(true, Ordering::SeqCst) {
+            die();
+        }
+
+        // TODO(eliza): abort any in-flight DMAs.
+
         // Ugly but works
         let mut uart: Uart = unsafe { core::mem::transmute(()) };
 
         write!(&mut uart, "\r\n").ok();
```
jamesmunns, line 284: You might want a null BEFORE as well, to force the end of the previous frame.

Additionally, you also probably want to either busy wait until the DMAC transfer is complete, or hard-stop any DMAC transfers that are ongoing, before manually driving the panic, otherwise I dunno how pushing bytes into the fifo while the DMAC is also doing it would go.

jamesmunns (approved):
Nits are TIOLI, and feel free to defer any of the nice to haves for later PRs if you want to merge now.

platforms/allwinner-d1/core/src/lib.rs
hawkw: yeah, i left a TODO about killing DMAC transfers. i intend to do that in a follow-up but will have to read the data sheet first.

PR #118, opened 2023-07-01
feat(allwinner-d1): cancel in-flight DMA on panic

hawkw:
This commit updates the Allwinner D1 panic handler to cancel all in-flight DMA requests when panicking.


jamesmunns (approved):
Awesome, looks great!

PR #122, opened 2023-07-02
feat(kernel): Add sermux shell daemon

jamesmunns:
This PR adds a helper daemon for spawning a shell on a sermux port

### DIFF
--- a/source/kernel/src/daemons/shells.rs
+++ b/source/kernel/src/daemons/shells.rs
@@ -146,7 +146,7 @@ impl GraphicalShellSettings {
     }
 }
 
-/// Spawns a graphical shell using the [EmbDisplayService][crate::services::emb_display::EmbDisplayService] service
+/// Spawns a graphical shell using the [EmbDisplayService](crate::services::emb_display::EmbDisplayService) service
 #[tracing::instrument(skip(k))]
 pub async fn graphical_shell_mono(k: &'static Kernel, settings: GraphicalShellSettings) {
     let GraphicalShellSettings {
--- a/source/kernel/src/daemons/shells.rs
+++ b/source/kernel/src/daemons/shells.rs
@@ -1,11 +1,19 @@
+//! Shells
+//!
+//! This module provides daemons that serve a forth shell
+
 use core::time::Duration;
 
 use crate::{
-    services::{emb_display::EmbDisplayClient, serial_mux::PortHandle},
+    forth::Params,
+    services::{
+        emb_display::EmbDisplayClient,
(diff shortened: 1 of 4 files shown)


hawkw (approved):
the fact that a shell is a "daemon" feels like a weird artifact of our naming scheme, but this PR looks good to me!

source/kernel/src/daemons/shells.rs
```diff
@@ -19,43 +27,150 @@ use profont::PROFONT_12_POINT;
...
+            forth_settings: Default::default(),
+            disp_width_px: width_px,
+            disp_height_px: height_px,
+            font: PROFONT_12_POINT,
+            _priv: (),
+        }
+    }
+}
+
+/// Spawns a graphical shell using the [EmbDisplayService][crate::services::emb_display::EmbDisplayService] service
```
hawkw, line 138: i think the rustdoc link syntax is wrong here, and this should be:
```suggestion
/// Spawns a graphical shell using the [EmbDisplayService](crate::services::emb_display::EmbDisplayService]) service
```

source/kernel/src/daemons/shells.rs
```diff
@@ -19,43 +27,150 @@ use profont::PROFONT_12_POINT;
...
+                }
+            }
+        }
+    })
+    .await;
+}
+
+/// Settings for the [graphical_shell_mono] daemon
+///
+/// This does NOT implement [Default]. Instead use [GraphicalShellSettings::with_display_size].
```
hawkw, line 97: maybe worth showing an example of this initializer:
```suggestion
/// This does NOT implement [Default]. Instead use [GraphicalShellSettings::with_display_size].
///
/// For example:
/// ```
/// use kernel::daemons::shells::GraphicalShellSettings;
/// let shell = GraphicalShellSettings {
///     // override the capacity with a larger value:
///     capacity: 512,
///    ..GraphicalShellSettings::with_display_size(420, 69), // nice!
/// };
/// # drop(shell);
/// ```
```

source/kernel/src/services/serial_mux.rs
```diff
@@ -40,6 +40,21 @@ pub enum WellKnown {
     PsuedoKeyboard = 2,
     /// A bidirectional for binary encoded tracing messages
     BinaryTracing = 3,
+
+    /// An interactive forth shell (1/4)
```
hawkw, line 44: nit, take it or leave it: it could be worth documenting the difference between `PseudoKeyboard` (just input, used for the graphical shell) and `ForthShell{N}` (which is bidirectional)?

PR #123, opened 2023-07-02
feat(sermux + crowtty): Move sermux/crowtty protocol into separate crate

jamesmunns:
This PR moves the sermux protocol format and definitions into a separate crate, removing the magic values from the crowtty crate

### DIFF
--- a/source/sermux-proto/src/lib.rs
+++ b/source/sermux-proto/src/lib.rs
@@ -1,0 +1,263 @@
+//! # sermux-proto
+//!
+//! Wire types used by the `SerialMuxService` in the kernel. Extracted as a
+//! separate crate to allow external decoders (like `crowtty`) to share protocol
+//! definitions
+
+#![cfg_attr(not(any(test, feature = "use-std")), no_std)]
+
+use core::mem::size_of;
+
+////////////////////////////////////////////////////////////////////////////////
+// Well Known Ports
+////////////////////////////////////////////////////////////////////////////////
+
+/// Well known [SerialMuxService] ports
+#[repr(u16)]
+#[non_exhaustive]
+pub enum WellKnown {
+    /// A bidirectional loopback channel - echos all characters back
+    Loopback = 0,
+    /// An output-only channel for sending periodic sign of life messages
+    HelloWorld = 1,
+    /// An input-only channel to act as a keyboard for a GUI application
+    /// such as a forth console, when there is no hardware keyboard available
+    PsuedoKeyboard = 2,
+    /// A bidirectional for binary encoded tracing messages
+    BinaryTracing = 3,
+}
+
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum Error {
+    /// The provided buffer is not suitable in size
+    InsufficientSize,
+    /// Ran out of room while filling a buffer, this is likely
+    /// an error in the `sermux-proto` library.
+    UnexpectedBufferFull,
+    /// The cobs decoding process failed. The message was likely
+    /// malformed or not a sermux-proto frame
+    CobsDecodeFailed,
+    /// Cobs decoding succeeded, but the resulting data was not
+    /// a valid sermux-proto frame
+    MalformedFrame,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct PortChunk<'a> {
+    pub port: u16,
+    pub chunk: &'a [u8],
+}
+
+impl<'a> PortChunk<'a> {
+    /// Create a new PortChunk from the given port and data
+    #[inline]
+    pub fn new(port: u16, chunk: &'a [u8]) -> Self {
+        Self { port, chunk }
+    }
+
+    /// Calculate the size required to encode the given data payload size
+    #[inline]
+    pub fn buffer_required(&self) -> usize {
+        // Room for COBS(port:u16 + data:[u8; len]) plus a terminating zero
+        cobs::max_encoding_length(self.chunk.len() + size_of::<u16>() + 1)
+    }
+
+    /// Encodes the current [PortChunk] into the given buffer
+    pub fn encode_to<'b>(&self, out_buf: &'b mut [u8]) -> Result<&'b mut [u8], Error> {
+        let PortChunk { port, chunk } = self;
+        if out_buf.len() < self.buffer_required() {
+            return Err(Error::InsufficientSize);
+        }
+
+        let mut encoder = cobs::CobsEncoder::new(out_buf);
+        encoder
+            .push(&port.to_le_bytes())
+            .map_err(|_| Error::UnexpectedBufferFull)?;
+        encoder
+            .push(chunk)
+            .map_err(|_| Error::UnexpectedBufferFull)?;
+        let used = encoder
+            .finalize()
+            .map_err(|_| Error::UnexpectedBufferFull)?;
+        // Get the encoded amount, with room for an extra zero terminator
+        let res = out_buf
+            .get_mut(..used + 1)
+            .ok_or(Error::UnexpectedBufferFull)?;
+        res[used] = 0;
+        Ok(res)
+    }
+
+    /// Decodes a [PortChunk] from the given buffer
+    ///
+    /// NOTE: This MAY mutate `data`, even if the decoding fails.
+    pub fn decode_from(data: &'a mut [u8]) -> Result<Self, Error> {
+        let dec_len = cobs::decode_in_place(data).map_err(|_| Error::CobsDecodeFailed)?;
+
+        // Messages must have a port and at least one data byte to be
+        // well formed
+        if dec_len < (size_of::<u16>() + 1) {
+            return Err(Error::MalformedFrame);
+        }
+
+        let frame = data.get(..dec_len).ok_or(Error::MalformedFrame)?;
+
+        let mut port_bytes = [0u8; size_of::<u16>()];
+        let (port_data, chunk) = frame.split_at(size_of::<u16>());
+        port_bytes.copy_from_slice(port_data);
+        let port = u16::from_le_bytes(port_bytes);
+
+        Ok(PortChunk { port, chunk })
+    }
+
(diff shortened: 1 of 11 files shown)


jamesmunns:
Note: This is going to conflict with #122 because I didn't plan ahead well. I'll fix the merge conflicts on whichever PR merges second.

source/sermux-proto/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+#[derive(Debug, PartialEq)]
+pub struct PortChunk<'a> {
+    pub port: u16,
+    pub chunk: &'a [u8],
+}
+
+impl<'a> PortChunk<'a> {
+    /// Create a new PortChunk from the given port and data
+    #[inline]
+    pub fn new(port: u16, chunk: &'a [u8]) -> Self {
```
hawkw, line 54: tiny nit, but it's nice to slap this on constructors
```suggestion
    #[must_use]
    pub fn new(port: u16, chunk: &'a [u8]) -> Self {
```

source/sermux-proto/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+        Self { port, chunk }
+    }
+
+    /// Calculate the size required to encode the given data payload size
+    #[inline]
+    pub fn buffer_required(&self) -> usize {
```
hawkw, line 60:
```suggestion
    #[must_use]
    pub fn buffer_required(&self) -> usize {
```

source/sermux-proto/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+        let port = u16::from_le_bytes(port_bytes);
+
+        Ok(PortChunk { port, chunk })
+    }
+
+    /// Convert into an [OwnedPortChunk]
+    ///
+    /// Only available with the `use-std` feature active
+    #[cfg(feature = "use-std")]
+    pub fn into_owned(self) -> OwnedPortChunk {
```
hawkw, line 116: nit/TIOLI: it seems like this could be an implementation of the `std::borrow::ToOwned` trait?

source/sermux-proto/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+        let (port_data, chunk) = frame.split_at(size_of::<u16>());
+        port_bytes.copy_from_slice(port_data);
+        let port = u16::from_le_bytes(port_bytes);
+
+        Ok(PortChunk { port, chunk })
+    }
+
+    /// Convert into an [OwnedPortChunk]
+    ///
+    /// Only available with the `use-std` feature active
```
hawkw, line 114: AFAICT this really only requires `alloc`, and not `std`? could be worth a separate feature...

source/sermux-proto/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+#[cfg(feature = "use-std")]
+pub struct OwnedPortChunk {
+    pub port: u16,
+    pub chunk: Vec<u8>,
+}
+
+#[cfg(feature = "use-std")]
+impl OwnedPortChunk {
+    /// Create a new PortChunk from the given port and data
+    #[inline]
```
hawkw, line 136:
```suggestion
    #[inline]
    #[must_use]
```

source/sermux-proto/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+/// Only available with the `use-std` feature active
+#[cfg(feature = "use-std")]
+pub struct OwnedPortChunk {
+    pub port: u16,
+    pub chunk: Vec<u8>,
+}
+
+#[cfg(feature = "use-std")]
+impl OwnedPortChunk {
+    /// Create a new PortChunk from the given port and data
```
hawkw, line 135:
```suggestion
    /// Create a new OwnedPortChunk from the given port and data
```

source/sermux-proto/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    #[inline]
+    pub fn new(port: u16, chunk: Vec<u8>) -> Self {
+        Self { port, chunk }
+    }
+
+    /// Calculate the size required to encode the given data payload size
+    #[inline]
```
hawkw, line 142:
```suggestion
    #[inline]
    #[must_use]
```

source/sermux-proto/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+        let mut data = data.to_vec();
+        let pc = PortChunk::decode_from(&mut data)?;
+        let port = pc.port;
+        let len = pc.chunk.len();
+        data.shrink_to(len);
+        Ok(OwnedPortChunk { port, chunk: data })
+    }
+
+    /// Borrows self as a [PortChunk]
+    pub fn as_port_chunk(&self) -> PortChunk<'_> {
```
hawkw, line 167: nit: we could probably have a `Borrow` impl that does this?

tools/crowtty/src/main.rs
```diff
@@ -286,50 +298,51 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                    success = true;
+
+                    // If the malformed frame is JUST a null terminator, this is probably
+                    // a "frame flush" event, like we are just about to panic.
+                    if carry != &[0x00] {
+                        println!("{tag} {dmux} {err} bonus data? {carry:#02x?}");
+                    }
                 }
-            } else {
+                Err(_) => {}
```
hawkw, line 338: should we hang onto the error in this case and print it?

tools/crowtty/src/main.rs
```diff
@@ -286,50 +298,51 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-                for line in s.lines() {
-                    println!("{tag} {text} {line}");
+                Err(SPError::MalformedFrame) => {
+                    success = true;
+
+                    // If the malformed frame is JUST a null terminator, this is probably
+                    // a "frame flush" event, like we are just about to panic.
+                    if carry != &[0x00] {
+                        println!("{tag} {dmux} {err} bonus data? {carry:#02x?}");
+                    }
```
hawkw, line 336: nit, not actually a huge deal: perhaps the rule that a single null byte is acceptable and should just be ignored could be encoded in the `sermux-proto` crate --- we could have it return a different error when decoding a single null byte?

source/sermux-proto/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// Ran out of room while filling a buffer, this is likely
+    /// an error in the `sermux-proto` library.
+    UnexpectedBufferFull,
+    /// The cobs decoding process failed. The message was likely
+    /// malformed or not a sermux-proto frame
+    CobsDecodeFailed,
+    /// Cobs decoding succeeded, but the resulting data was not
+    /// a valid sermux-proto frame
+    MalformedFrame,
+}
```
hawkw, line 43: this isn't a huge deal, but: the `InsufficientSize` and `UnexpectedBufferFull` errors are returned only when encoding, and `CobsDecodeFailed` and `MalformedFrame` errors are returned only when encoding --- maybe it's worth splitting this type into separate `EncodeError` and `DecodeError` enums, so that when handling the result of an encode or decode operation, the user doesn't have to match on variants that are never returned by that operation?

source/sermux-proto/src/lib.rs
hawkw, line 54: nit, not a big deal: if we changed this to 
```suggestion
    pub fn new(port: impl Into<u16>, chunk: &'a [u8]) -> Self {
```
and added an `impl Into<u16> for WellKnown`, you could write
```rust
PortChunk::new(WellKnown::HelloWorld, b"hello world")
```
without having to cast the well known port enum?

not actually important!

source/sermux-proto/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum Error {
+    /// The provided buffer is not suitable in size
+    InsufficientSize,
+    /// Ran out of room while filling a buffer, this is likely
+    /// an error in the `sermux-proto` library.
+    UnexpectedBufferFull,
+    /// The cobs decoding process failed. The message was likely
+    /// malformed or not a sermux-proto frame
+    CobsDecodeFailed,
```
hawkw, line 39: nit, take it or leave it: should this wrap the error type returned by the `cobs` crate?

source/sermux-proto/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// An output-only channel for sending periodic sign of life messages
+    HelloWorld = 1,
+    /// An input-only channel to act as a keyboard for a GUI application
+    /// such as a forth console, when there is no hardware keyboard available
+    PsuedoKeyboard = 2,
+    /// A bidirectional for binary encoded tracing messages
+    BinaryTracing = 3,
+}
+
+#[derive(Debug, PartialEq, Copy, Clone)]
```
hawkw, line 30: nit: would be nice if this had a `Display` impl as well

hawkw (approved):
this is awesome! i left a bunch of very minor suggestions, but none of them are terribly important.

hawkw:
oh, we might also want to add `sermux-proto` to the list of crates in `source/`, here: https://github.com/tosc-rs/mnemos/tree/main/source#folder-layout

source/sermux-proto/src/lib.rs
jamesmunns: The cobs error type is `()`

source/sermux-proto/src/lib.rs
jamesmunns: happy to add an alloc-only feature if we ever need it

source/sermux-proto/src/lib.rs
hawkw: well, nevermind then!

PR #125, opened 2023-07-04
Remove some residual testing code

jamesmunns:
I uh, left this in while testing the panic code in #123. It generally shouldn't do that.


hawkw (approved):
whoopsie

PR #132, opened 2023-07-06
feat: janky way of having tracing in ISRs

hawkw:
This is not *that* good but it allows at least one ISR to trace even if
the interrupt fired while normal code was holding the trace WGR. We can
probably figure out a nicer version of this, with different tracebufs
for each ISR priority level, later.

Cherry-picked out of #103

### DIFF
--- a/platforms/allwinner-d1/core/src/lib.rs
+++ b/platforms/allwinner-d1/core/src/lib.rs
@@ -234,7 +234,6 @@ impl D1 {
     /// At the moment, we only service the Channel 0 interrupt,
     /// which indicates that the serial transmission is complete.
     fn handle_dmac() {
-        let _isr = kernel::isr::Isr::enter();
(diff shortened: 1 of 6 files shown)


platforms/allwinner-d1/core/src/lib.rs
```diff
@@ -234,8 +234,10 @@ impl D1 {
     /// At the moment, we only service the Channel 0 interrupt,
     /// which indicates that the serial transmission is complete.
     fn handle_dmac() {
+        let _isr = kernel::isr::Isr::enter();
```
jamesmunns, line 237: We could potentially do this at the top of the actual ISR entry point instead of in every handler

hawkw: done!

jamesmunns (approved):
shipit.swf

PR #135, opened 2023-07-07
feat(crowtty): basic built-in pseudo-keyboard

hawkw:
Currently, the SerMux pseudo-keyboard port is handled like any other SerMux port by crowtty --- so, if we want to send keyboard input to the target device, we must open a separate `stty`/`netcat` session to write to the pseudo-keyboard. This is a bit annyoing and requires opening an additional terminal window.

To improve developer quality of life, I've added a very basic built-in pseudo-keyboard to crowtty. Now, input from stdin is written to the pseudo-keyboard port by default. A command-line argument allows disabling this behavior and treating the pseudo-keyboard port as a normal SerMux port, so the old behavior is still available.

The current pseudo-keyboard is pretty basic. It sends input to the target line-by-line rather than character-by-character, so we can't see typing on the target's display. Also, everything written to the port is displayed on the terminal on the host, which is a bit ugly when it gets interrupted by a log line. In the future, we may want to use a terminal UI library to change the behavior for stdin to read character-by-character and not echo the input, so that the input is instead echoed on the target's display as the user types it in. But this works for now.

Also, I fixed a typo in the `sermux_proto::WellKnown` enum --- it's spelt "pseudo", not "psuedo". :)

For example: 
image

And on the target:
image

### DIFF
--- a/source/kernel/src/daemons/shells.rs
+++ b/source/kernel/src/daemons/shells.rs
@@ -108,9 +108,9 @@ pub async fn sermux_shell(k: &'static Kernel, settings: SermuxShellSettings) {
 /// ```
 #[derive(Debug)]
 pub struct GraphicalShellSettings {
-    /// Sermux port to use as a PsuedoKeyboard.
+    /// Sermux port to use as a PseudoKeyboard.
     ///
-    /// Defaults to [WellKnown::PsuedoKeyboard]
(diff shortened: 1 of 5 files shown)


jamesmunns (approved):
Generally cool with this!

PR #148, opened 2023-07-12
add web-based simulator

spookyvision:
this PR adds a web-based (WASM) simulator, codenamed "pomelo". 
It features an interactive shell which allows you to
- start (but not stop) the hello service
- interact with the loopback service via the `echo` command
- interact with a forth process using `forth <single expression>` or `iforth` for a REPL.

The shell includes a command history that can be navigated with cursor up/down, and (very) basic tab completion.

Pomelo harnesses `wasm-tracing` to send tracing messages to the browser debug console. Chrome can further use the tracing spans to show a profiling graph.


hawkw:
sooooo cool!

source/kernel/Cargo.toml
```diff
@@ -80,6 +80,7 @@ path = "../spitebuf"
 
 [dependencies.mnemos-alloc]
 version = "0.1.0"
+path = "../alloc"
```
jamesmunns, line 83: fwiw: I think this is normally addressed by the top level workspace patch. You could instead add this to your pomelo workspace overrides

jamesmunns (approved):
Looks good! Let's get this on the road :)

PR #154, opened 2023-07-13
feat(alloc): add some more methods to `FixedVec`

hawkw:
These seemed nice to have. Some of them were needed for implementing the keyboard driver.


jamesmunns:
Nice! Would you mind also swapping https://github.com/tosc-rs/mnemos/blob/main/source/melpomene/src/sim_drivers/emb_display.rs#L301-L317 to use the new method?

jamesmunns (approved):
Looks good! We could link to the alloc vec docs for these methods, but also fine not to.

PR #155, opened 2023-07-13
feat(beepy): add a mediocre `i2c_puppet` driver

hawkw:
Depends on #154.

This branch adds a simple driver for the `i2c_puppet` firmware on the
Beepy RP2040. It allows controlling the status LED and reading
keypresses.

Key input can be read by sending a message to the driver service to
subscribe to keypresses. The driver task tracks a list of subscriptions
(channels) and broadcasts each keystroke to all subscribers. In
addition, the status LED and backlight can be controlled by sending
messages to the driver task.

The keyboard driver is in a new `mnemos-beepy` crate in `platforms` that
contains drivers for the Beepy hardware that are implemented using
cross-platform driver service definitions. Right now, this is just the
`i2c_puppet` driver, which is implemented against the kernel's
cross-platform `I2cService` interface. However, in the future, if we can
write a generic driver interface for SPI in the kernel crate, we could
move the SHARP display driver into `mnemos-beepy` as well.

Currently, the way the keyboard is implemented is pretty bad: we poll
the keyboard for new keystrokes periodically, as long as at least one
task is subscribed. This is wildly inefficient and we should instead
have the `i2c_puppet` send us an interrupt when new keystrokes are
ready. However, this will require code for configuring the D1 to
actually generate interrupts on GPIO pins, which we haven't written yet.
I'll add this in a later branch and rewrite the keyboard driver to use
that.

Keyboard input isn't actually wired up to anything yet --- we just run a
task that logs all keypresses. A subsequent branch will actually make
the forth shell handle keyboard input.


platforms/allwinner-d1/boards/Cargo.toml
```diff
@@ -72,3 +75,8 @@ rev = "cbcfc62a6ea3646fb43f2c159cfdc19b3d932004"
 
 [patch.crates-io.mnemos-alloc]
 path = "../../../source/alloc"
+
+
+[patch.crates-io.bbq10kbd]
```
jamesmunns, line 80: Maybe add a note to keep in sync with the other patch at the top level?

platforms/allwinner-d1/core/src/drivers/twi.rs
```diff
@@ -349,8 +345,9 @@ impl I2c0 {
         let mut guard = self.isr.lock(self.twi);
 
         // reset the TWI engine.
-        guard.twi.twi_srst.write(|w| w.soft_rst().set_bit());
-        guard.twi.twi_efr.reset();
+        // guard.twi.twi_srst.write(|w| w.soft_rst().set_bit());
```
jamesmunns, line 348: Remove dead code?

platforms/beepy/src/i2c_puppet.rs
```diff
@@ -0,0 +1,632 @@
...
+        let subscriptions = FixedVec::new(settings.max_subscriptions).await;
+        {
+            // first, make sure we can get the version...
+            let mut kbd = AsyncBbq10Kbd::new(&mut i2c);
+            let Version { major, minor } = kbd
+                .get_version()
+                .await
+                .map_err(RegistrationError::NoI2cPuppet)?;
+            tracing::info!("i2c_puppet firmware version: v{major}.{minor}");
+            // // and then, reset i2c_puppet
```
jamesmunns, line 261: More dead code

jamesmunns: And if we *don't* remove this, might want to check what the i2c_puppet firmware does on reset: technically it switches the 5v power that normally powers the mango pi (when we actually start using battery power).

We want to make sure we don't reset IT and then it resets US.

jamesmunns (approved):
Couple of nits, overall looks good! Looking forward to playing with it :D

platforms/beepy/src/i2c_puppet.rs
hawkw: okay yeah we don't want that

PR #156, opened 2023-07-13
feat(crowtty): add host-side trace filtering

### LINKED ISSUE #153: crowtty should have client side trace filtering
on targets/mod paths. this would make my life way easier

hawkw:
Currently, our ability to filter traces is limited to setting a max
level on the target. This is not ideal --- sometimes, we may want
`TRACE` events from one module, but enabling `TRACE` globally results in
an unreadably verbose firehose log.

In the rest of the `tracing` ecosystem, this is generally solved by
using *filters*, typically setting a separate maximum level based on
trace targets (module paths). We can't easily implement dynamic target
filtering in the kernel itself, because `mnemos_alloc` doesn't currently
provide us a nice way of working with owned strings for mudle paths.

However, we *can* easily implement this kind of filtering on the host
side, in `crowtty`. This branch adds support for module-path-based
filtering in `crowtty`, controlled by a `--trace` flag, which replaces
the previous `--trace-level`. I felt like breaking the CLI arguments was
probably fine but if someone really cares about backwards compatibility,
we could make the `--trace-level` flag an alias for `--trace <LEVEL>`.

In the future, when we have the capacity to do nice string handling in
the kernel, it would be preferable to have these trace filter
configurations be sent to the target device, and have it perform
filtering. That would allow us to actually skip recording unwanted
traces, reducing the performance impact of turning on more verbose
traces. However, with the current design, we will still send the max
level to the target, so if we enable `INFO` globally and `DEBUG` in some
specific module, we will, at least, still disable the `TRACE` level on
the target.

When the capacity to perform filtering in the target is added, the
existing flag can be kept, and the behavior modified to send a
traceproto message that selects that filter configuration.

For example, here's a `crowtty` session where I enable the `TRACE` level
for the TWI driver, but `INFO` for all other modules. Note that none of
the traces from other parts of mnemOS, such as `comms::bbq` or the SHARP
display driver, are displayed, so we only see the very verbose
diagnostics from the TWI driver.
image

Fixes #153

### DIFF
--- a/tools/crowtty/src/main.rs
+++ b/tools/crowtty/src/main.rs
@@ -70,12 +70,12 @@ struct Args {
     /// - the `INFO` level globally (regardless of module path),
     /// - the `DEBUG` level for all modules in the `kernel` crate,
     /// - and the `TRACE` level for the `comms::bbq` submodule in `kernel`.
-    /// 
+    ///
     /// enabling a more verbose level enables all levels less verbose than that
     /// level. for example, enabling the `INFO` level for a given target will also
-    /// enable the `WARN` and `ERROR` levels for that target. 
+    /// enable the `WARN` and `ERROR` levels for that target.
     ///
-    /// see <https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/targets/struct.Targets.html#filtering-with-targets> 
+    /// see <https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/targets/struct.Targets.html#filtering-with-targets>
     /// for more details on this syntax.
     #[arg(
         short,
--- a/tools/crowtty/src/main.rs
+++ b/tools/crowtty/src/main.rs
@@ -63,9 +63,15 @@ struct Args {
     #[arg(short, long, global = true)]
     verbose: bool,
 
-    /// maximum `tracing` level to request from the target.
-    #[arg(short, long, global = true, default_value_t = LevelFilter::INFO)]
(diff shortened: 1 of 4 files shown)


tools/crowtty/src/main.rs
```diff
@@ -63,9 +63,15 @@ struct Args {
     #[arg(short, long, global = true)]
     verbose: bool,
 
-    /// maximum `tracing` level to request from the target.
-    #[arg(short, long, global = true, default_value_t = LevelFilter::INFO)]
-    trace_level: LevelFilter,
+    /// a comma-separated list of `tracing` targets and levels to enable.
```
jamesmunns, line 66: It might be nice to have an example of how to use this in the CLI output, or maybe in the crowtty readme (does crowtty have a readme?)? Maybe something like `--trace=kernel=log,maitake=trace,postcard::flavors=debug` or whatever.

jamesmunns (approved):
One docs nit, but no complaints and I'm not worried about breaking cli arg changes atm.

tools/crowtty/src/main.rs
hawkw: crowtty has basically no docs lol...we should fix that.

hawkw: https://github.com/tosc-rs/mnemos/pull/156/commits/f5b52500034a90db9df8ac1483d2f1a86a35a88d adds a quick description of how this works.

PR #162, opened 2023-07-16
feat(emb_display): Introduce version two of the `EmbDisplay` service

jamesmunns:
This PR deprecates the old `EMB_DISPLAY` driver, and introduces `EMB_DISPLAY_V2`.

The new design removes the generally-not-actually-a-thing expectation that frames should be allocated and managed by the display driver, and instead callers are expected to allocate "frame chunks" that they need.

Frame chunks are a subsection of the display, and also feature a "mask" that can be used to only draw a relevant subsection of the chunk to the display, similar to how blitting masks are used.

Frames chunks are also returned to the caller once they have been applied to the display (or a framebuffer managed by the display), meaning they can be re-used across multiple draw calls.

Melpomene and the d1 sharp display drivers have been updated to use the v2 API.

Additionally, the D1's SPI implementation has been updated to draw less of the screen less often when not necessary, and also now uses a waitqueue, due to issues found with racing the WaitCell in small transfers.

Proof of still working:

https://github.com/tosc-rs/mnemos/assets/2097964/2838b08f-35d7-42bb-a2b1-297bc329f025


source/alloc/src/containers.rs
```diff
@@ -275,6 +275,106 @@ impl<T> DerefMut for ArrayBuf<T> {
...
+/// own safety invariants, without manually using `alloc`.
+pub struct HeapArray<T> {
+    ptr: NonNull<T>,
+    len: usize,
+}
+
+unsafe impl<T: Send> Send for HeapArray<T> {}
+unsafe impl<T: Sync> Sync for HeapArray<T> {}
+
+impl<T> HeapArray<T> {
```
jamesmunns, line 297: Note to self: There's a couple places we could replace `FixedVec` with this.

jamesmunns:
This is generally ready for review now, however I haven't yet updated the D1 hardware implementations.

source/kernel/src/registry/mod.rs
```diff
@@ -451,7 +460,29 @@ impl Registry {
...
+            body: contents,
+            service_id: self.service_id,
+            client_id: self.client_id,
+            request_id: self.request_id,
+        }
+    }
+}
+
 impl<P> Envelope<P> {
+    // TODO: constrain types
```
jamesmunns, line 475: I can remove this, it's now private and `Message::split` does constrain types appropriately.

source/melpomene/src/sim_drivers/emb_display.rs
```diff
@@ -89,22 +89,27 @@ impl SimDisplay {
...
     dirty: bool,
 }
 
 impl CommanderTask {
     /// The entrypoint for the driver execution
-    async fn run(mut self, width: u32, height: u32) {
+    async fn run(self, width: u32, height: u32) {
         let output_settings = OutputSettingsBuilder::new()
             .theme(BinaryColorTheme::OledBlue)
+            .scale(1)
```
jamesmunns, line 108: I should remove this, or we should make it configurable. With hidpi on my mac, the default scale of 2 is huge, where it wasn't quite as huge on linux with no hidpi scaling.

source/melpomene/src/sim_drivers/emb_display.rs
```diff
@@ -129,186 +136,170 @@ impl CommanderTask {
...
+        let mut guard = mutex.lock().await;
+        let mut done = false;
+        if let Some(Context {
+            sdisp,
+            window,
+            dirty,
+            ..
+        }) = (&mut *guard).as_mut()
+        {
+            // If nothing has been drawn, only update the frame at 5Hz to save
```
jamesmunns, line 216: I was drawing at 1hz before, but it made MacOS sad and it would make the frame blink, likely greying it out when it thought the window was "unresponsive". I should look if there's an "update but don't redraw" option.

source/melpomene/src/sim_drivers/emb_display.rs
```diff
@@ -129,186 +136,170 @@ impl CommanderTask {
...
+        let src = src_data.iter().zip(src_mask.iter());
+
+        dest_row
+            .iter_mut()
+            // Skip to the start of the subframe
+            .skip(start_x as usize)
+            // Again, zipping means we stop as soon as we run out of
+            // source OR destination pixesl on this line
+            .zip(src)
+            .filter_map(|(d, (s_d, s_m))| {
```
jamesmunns, line 278: Oh I forgot to add my bit hax version here, I should add it.

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
hawkw, line 1: not really related at all to this PR but i feel like this ought to eventually move to the `mnemos-beepy` crate instead of `mnemos-d1`...but that depends on having a SPI service interface in the kernel crate probly?

hawkw: anyway not directly relevant to this PR at all

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -24,20 +24,18 @@
 //!
 //! Reference: <https://www.sharpsde.com/fileadmin/products/Displays/2016_SDE_App_Note_for_Memory_LCD_programming_V1.3.pdf>
 
-use core::time::Duration;
+use core::{convert::identity, time::Duration};
```
hawkw, line 27: i love any code that actually has a reason to use `core::convert::identity` :D

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -132,12 +123,14 @@ impl SharpDisplay {
 /// One entire frame, stored one bit per pixel
 struct FullFrame {
     frame: [[u8; WIDTH_BYTES]; HEIGHT],
+    dirty_line: [bool; HEIGHT],
```
hawkw, line 126: tiny nit: this naming is a little unclear to me, i would have maybe called it `dirty_lines` to make it clear that it's a map of _all_ lines that might be dirty. looking at the code below where it's referenced i mistakenly thought we were iterating over the pixels of a single dirty line, at first.

hawkw: Side Thing, i wonder if we could golf this into using a bitmap instead of a bytemap. probably not actually important as we have a whopping Entire Gigabyte of RAM...

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -247,35 +208,64 @@ impl Draw {
...
+                // Filter out to only the dirty lines, clearing the dirty flag
+                let all_lines = dirty_line.iter_mut().zip(frame.iter()).enumerate();
+                let dirty_lines = all_lines.filter_map(|(idx, (dirty, line))| {
+                    if *dirty {
+                        *dirty = false;
+                        Some((idx, line))
+                    } else {
+                        None
+                    }
+                });
```
hawkw, line 238: this is cute!

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -247,35 +208,64 @@ impl Draw {
...
+
+                // Now we need to write all the dirty lines, zip together the dest buffer
+                // with our current frame buffer
+                for (line, iline) in dirty_lines {
+                    // Lines are 1-indexed on the wire
+                    let _ = self.buf.try_push((line as u8) + 1);
+                    // We keep our internal frame buffer in the same format as the wire
+                    let _ = self.buf.try_extend_from_slice(iline);
+                    // dummy byte
+                    let _ = self.buf.try_push(0);
```
hawkw, line 248: nit/TIOLI: maybe the `try_push`/`try_extend_from_slice` calls should be `expect`ed? if the buffer ever doesn't have capacity we fucked up kinda bad, since it's preallocated to the max capacity we need...

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
```diff
@@ -390,82 +346,51 @@ struct Context {
...
-            width,
-            height,
-        };
+    let s = data
+        .chunks(src_width as usize)
+        .zip(mask.chunks(src_width as usize));
 
-        Ok(fc)
-    }
+    let before = dest.dirty_line.iter().filter(|b| **b).count();
```
hawkw, line 371: nit, take it or leave it: it might be worth guarding this behind [`event_enabled!`](https://docs.rs/tracing/latest/tracing/macro.event_enabled.html), since it's only used if `trace!` is enabled, and we could avoid iterating if we're not going to log the trace event?
```suggestion
    let before = if tracing::event_enabled!(tracing::Level::TRACE) {
         dest.dirty_line.iter().filter(|b| **b).count()
    } else {
        0
    };
``` 
this is probably a premature optimization though :woman_shrugging:

source/kernel/src/daemons/shells.rs
```diff
@@ -143,7 +144,8 @@ impl GraphicalShellSettings {
 }
 
 /// Spawns a graphical shell using the [EmbDisplayService](crate::services::emb_display::EmbDisplayService) service
-#[tracing::instrument(skip(k))]
+// TODO: tracing the `settings` field draws the whole PROFONT_12_POINT, which is hilarious but annoying
```
hawkw, line 147: thanks for removing that lol

source/kernel/src/daemons/shells.rs
```diff
@@ -214,83 +220,103 @@ pub async fn graphical_shell_mono(k: &'static Kernel, settings: GraphicalShellSe
...
-        loop {
-            // loop *just* for skipping released key events
-            futures::select_biased! {
-                event = keyboard.next().fuse() => {
+        let _ = k
+            .timeout(
+                interval,
+                io_poll(&mut any, &mut keyboard, &mut rline, &tid_io),
+            )
+            .await;
```
hawkw, line 247: it might be good to have a bigger comment here explaining the debouncing timeout, and why it's okay to cancel `io_poll` abruptly if it times out?

source/kernel/src/daemons/shells.rs
```diff
@@ -214,83 +220,103 @@ pub async fn graphical_shell_mono(k: &'static Kernel, settings: GraphicalShellSe
...
-                            rline.submit_remote_editing();
-                        } else {
-                            let _ = rline.append_remote_char(b);
-                        }
+            },
+            output = tid_io.consumer().read_grant().fuse() => {
+                let len = output.len();
+                tracing::trace!(len, "Received output from tid_io");
+                for &b in output.iter() {
+                    // TODO(eliza): what if this errors lol
```
hawkw, line 309: :)

source/kernel/src/registry/mod.rs
```diff
@@ -167,6 +171,13 @@ pub struct Envelope<P> {
     request_id: RequestResponseId,
 }
 
+pub struct OpenEnvelope<P> {
```
hawkw, line 174: a doc comment explaining this might be nice?
also, should probably be
```suggestion
#[must_use]
pub struct OpenEnvelope<P> {
```

source/kernel/src/registry/mod.rs
```diff
@@ -455,7 +466,29 @@ impl Registry {
 
 // Envelope
 
+impl<P> OpenEnvelope<P> {
+    pub fn fill(self, contents: P) -> Envelope<P> {
```
hawkw, line 470: nit, take it or leave it: is there a reason we can't get rid of the `P` generic on `OpenEnvelope` and just have 
```suggestion
impl OpenEnvelope {
    pub fn fill<P>(self, contents: P) -> Envelope<P> {
```
this might reduce some unneeded monomorphization possibly?

source/kernel/src/services/emb_display.rs
```diff
@@ -137,81 +103,244 @@ impl EmbDisplayClient {
...
+/// overwriting the existing content or background outside the circle.
+///
+/// This could also be used to keep persistently drawn [sprites] in memory,
+/// layering them onto the target frame.
+///
+/// All FrameChunk kinds have some kind of transparency, though this may range from
+/// a single bit transparency (transparent or not), to a more complex [alpha channel]
+///
+/// [alpha channel]: https://en.wikipedia.org/wiki/Alpha_compositing
+/// [sprites]: https://en.wikipedia.org/wiki/Sprite_(computer_graphics)
```
hawkw, line 191: love these docs

source/kernel/src/services/emb_display.rs
```diff
@@ -137,81 +103,244 @@ impl EmbDisplayClient {
...
+// On the other hand, it is likely a bit less computationally intense to
+// stick with byte addressing, as we don't need to do shifting and such
+// for individual pixel operations, but there might be nice ways to accelerate
+// that, though then you need to worry about "alignment" of data, e.g. if the
+// start_x is not a multiple of 8.
+//
+// It may just be worth adding a "MonoChunk1bpp" variant in the future to allow
+// users to make the size/perf tradeoff, particularly if we want to support
+// targets with very small memory. For example, a 400x240 monochrome display would
+// be 93.75KiB at 8bpp, but only 11.72KiB at 1bpp.
```
hawkw, line 217: taking a `bitvec` dep probably lets us do 1bpp without having to think too hard about How To Do Tricky Bit Operations, and @myrrlyn is cool...

source/kernel/src/services/emb_display.rs
```diff
@@ -222,24 +351,91 @@ impl DrawTarget for FrameChunk {
...
+
+#[derive(Copy, Clone)]
+pub struct FrameChunkMetadata {
+    start_x: u32,
+    start_y: u32,
+    width: u32,
+    height: u32,
+}
+
+/// Metadata of the [FrameChunk]
```
hawkw, line 418: tiny nit: i think you meant:
```suggestion
/// Metadata of the [FrameChunk]
#[derive(Copy, Clone)]
pub struct FrameChunkMetadata {
    start_x: u32,
    start_y: u32,
    width: u32,
    height: u32,
}

```

hawkw (approved):
very cool, props for lots of cute use of `zip`! left some unimportant nits on stuff, but this is great!

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
jamesmunns: Yeah, we definitely can! I make some notes about this in the code, it'd probably be a little more computationally intense to do all the bithax, but I ran out of brain juice to do it in this PR.

The hope is that `FrameChunk` can have more variants for different color depth in the future (and ideally we can convert across them), so we could have an 8bpp mono mode when you care more about perf, and a 1bpp mono mode when you care about size (and also 8/16/24/32bpp color modes).

source/kernel/src/services/emb_display.rs
jamesmunns: Yeah, I think I mentioned the same in chat, and meant to put it in the todo here, but I also think it's a very good idea. We mostly just need to make `mnemos_alloc::containers` versions of `BitVec` (or likely rather `BitArr`), and I think the bitvec crate has traits for that.

source/kernel/src/registry/mod.rs
jamesmunns: Gunna defer on this because I want to revisit this and move the whole API over to this to get rid of the old bad one.

source/kernel/src/registry/mod.rs
jamesmunns: Gunna defer on this because I want to revisit this and move the whole API over to this to get rid of the old bad one.

source/kernel/src/daemons/shells.rs
jamesmunns: Gunna defer on this because I hope to fix it once Instant is a thing. This should be handled in the display IMO and not the shell.

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
jamesmunns: Yeah, I'd probably pass since iterating over bools is probably going to optimize extremely well.

platforms/allwinner-d1/core/src/drivers/sharp_display.rs
jamesmunns: Yeah, probably. Too much embedded trauma with panic bloat makes me pretty wary of it though.

PR #163, opened 2023-07-16
feat(hw): Add mnemos shield

jamesmunns:
This PR adds the CAD design files for the second version of the mnemos shield. This design matches the version of the boards that were built for @hawkw.

The original revision had the center row of headers with 1.2mm PTH, which were too large to be functional for the pogo pins. This history is not included (because it didn't fully work).

Gerber exports, suitable for ordering from JLC or other board fabs will be added as a github release after this has been merged.


hardware/mnemos-shield/README.md
hawkw, line 1: should we maybe include some kind of note about how the 5v rails need to be removed/disabled due to current software limitations? i don't want people frying their SOCs...

hawkw (changes requested):
overall this is great! however, i do wonder if we ought to include a note about the need to physically isolate the 5v rails from the MQ Pro/Beepy?

hardware/mnemos-shield/README.md
jamesmunns: Sure, how far would you like me to go in documenting this?

The connection of the 5v rail is DESIRED when you are not powering the board through the top and bottom of the device at the same time.

We are in the awkward between time where we must have the device tethered to the primary USB cable for programming, and have no persistence.

I could describe all of these options, or I could just go redesign the board to make this a breakable connection, either using a DPDT switch or solder bridge.

In particular, we don't perform WORSE or differently from the existing beepy (you're not supposed to plug into the pi usb ports at all), we are just weird because we HAVE to for ephemeral programming.

hawkw (approved):
fine to merge this whenever you like btw :)

Issue #164, opened 2023-07-17
fix(melpo): put back melpo keyboard support


hawkw:
I think #158 may have broken the pseudokeyboard support in Melpomene, since the Forth shell now relies on the `KeyboardMux` service, which Melpomene doesn't currently spawn.

We should also consider adding a "physical keyboard" to Melpo and have keypresses sent to the TTY that Melpo is running in publish keyboard events to the keymux service.

Issue #160, opened 2023-07-15
Idea collector: Configuration System


jamesmunns:
## Abstract

We currently have a lot of magic numbers for things like buffer sizes, number of ports, etc. We also don't particularly provide any guidance for what these numbers should be. In the immediate future, it would be good to have one central "config" table that can be used across the system for setting (and viewing the settings of) these values, even if the values are "baked in" at compile time.

## Future Features

In the future, we might want to have other neat features, such as:

* Layering of configs, so you can say "gimme all the defaults BUT change/add these"
* Loading from a file from disk at boot, instead of only at compile time
* Potentially changing these settings at runtime, if we ever do/want to support hot-reloading of services, so you can edit a config file, and restart a service

I think feature wise we'll probably also want some kind of namespacing, so we can get specific config values, like "sermux outgoing frame size".

## Possible Approaches

I think there's a few main stylistic ways to go about this:

### Option 1/N

**Go all strings**, using a very simple key:value system, where all settings are basically `&[(&str, &str)]`. Probably the simplest to write, pushes all the parsing + validation down to the consumers of the settings. We could potentially take a JSON or TOML at compile time and convert it to the "settings array" with a build-rs or something.

example for sermux: 

```rust
cfg = [
    ("services/54c983fa-736f-4223-b90d-c4360a308647/max_frame", "512"),
    // ...
];
```

### Option 2/N

Keep simple string keys + array of tuples, but also **support a limited number of primitive "value" types**, such as `u8`, `i32`, `usize`, `str`, etc. Could probably be used mostly the same as Option 1, but gives a little more type safety, but still there is a potential for "we wanted a `u8` but got an `i32`, what do?".

example for sermux:

```rust
cfg = [
    ("services/54c983fa-736f-4223-b90d-c4360a308647/max_frame", CfgVal::usize(512)),
    // ...
];
```

### Option 3/N

**Allow for some kind of namespace schema**, possibly not for ALL namespaces, but for some. This might require adding a `Settings` parameter to the `RegisteredService` trait, which we've talked about informally before. I'm not sure how to do this best, but we could likely rely on serde (with JSON/TOML/YAML/Postcard backends) for decoding.

example for sermux: 

```rust
cfg.add_service_cfg<SerialMuxService>(SerialMuxSettings {
    max_ports: 16,
    max_frame: 512,
});
```

#### Open questions:

[...]

PR #165, opened 2023-07-17
feat(kernel): add `initialize_default_services`

hawkw:
One issue with the design of kernel services is that the binding between
services and their clients is not enforced at compile time. This means
that it's possible to add new cross-platform services which should be
spawned on all hardware platforms without the hardware platforms failing
to compile if those services are not spawned.

In order to avoid issues where we forget to spawn a cross-platform
service on a particular platform, such as #164, this branch adds a new
`Kernel::initialize_default_services` that spapwns all the default
cross-platform services and daemon tasks. Rather than having to
duplicate the code to spawn these services in each platform
implementation, we can now simple call this method in the platform impl.
New services which we expect to always be spawned can simply be added
here in the cross-platform kernel, and we don't have to remember to add
code for spawning them to each platform.

To allow the platform to customize settings for the default services,
this method takes a struct containing all the services'
`${SERVICE}Settings` types. In the future, that will probably be
replaced by #160.

While working on this, I also did a bit of cleanup/refactoring of the
various services' registration methods and `Settings` types to make them
a bit more consistent.

Closes https://github.com/tosc-rs/mnemos/issues/164

### DIFF
--- a/platforms/allwinner-d1/core/src/lib.rs
+++ b/platforms/allwinner-d1/core/src/lib.rs
@@ -47,8 +47,6 @@ pub struct D1 {
     i2c0_int: (Interrupt, fn()),
 }
 
-static COLLECTOR: trace::SerialCollector = trace::SerialCollector::new();
-
 impl D1 {
     /// Initialize MnemOS for the D1.
     ///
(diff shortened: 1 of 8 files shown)


jamesmunns (approved):
I'm okay with merging this, as it's fine *for now*, but I think trying to have one central "all services" list is not something that is super long term viable, and would be much better served as something configuration controlled at compile and/or runtime (e.g. something kconfig+systemctl service list... ish).

We don't HAVE anything like that (and #160 only addresses part of that), so it doesn't make sense to defer something that is better today for something theoretically "better-er" that doesn't exist.

Basically - +1 to merge, but I'd like to get away from this once it is reasonable to.

There are also a lot of dead imports and broken docs links, but feel free to merge once those are resolved.

PR #166, opened 2023-07-18
fix(trace): always reregister when crowtty connects

hawkw:
Currently, we only re-register all `tracing` callsites if we receive a host trace level request *and* the level has changed. This means that if crowtty disconnects and then a new crowtty session connects, and the new session requests the same max level as the previous one, we won't register the callsites again. This means that the new session won't be able to understand any trace spans/events we have previously sent metadata for.

This branch removes the check that the level has changed, and always blows away the cache. This causes `tracing` to re-register all callsites every time a new `crowtty` session connects, so that the new session can actually understand all trace events.

### DIFF
--- a/source/kernel/src/trace.rs
+++ b/source/kernel/src/trace.rs
@@ -162,21 +162,19 @@ impl SerialCollector {
                                 let level = lvl
                                     .map(|lvl| lvl as u8)
                                     .unwrap_or(level_to_u8(LevelFilter::OFF));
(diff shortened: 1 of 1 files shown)


jamesmunns (approved):
Oh hell yeah, I noticed this a couple times and didn't know why it was doing this. yay QoL fixes!

hawkw:
> Oh hell yeah, I noticed this a couple times and didn't know why it was doing this. yay QoL fixes!

it's because im dumb lol

PR #174, opened 2023-07-20
chore: add `just` recipes for `crowtty` and `melpomene`

hawkw:


jamesmunns (approved):
LGTM, do you want to yank `cargo melpo` now since it doesn't play totally nice with workspaces in some cases?

hawkw:
> LGTM, do you want to yank `cargo melpo` now since it doesn't play totally nice with workspaces in some cases?

yeah will do!

PR #183, opened 2023-07-24
WIP: Initial support for oranda

jamesmunns:
This generally works, but has not been wired up to CI, and probably won't be until the next release of oranda (with workspace support) lands.


oranda-workspace.json
```diff
@@ -0,0 +1,68 @@
+{
+  "workspace": {
+    "name": "MnemOS",
+    "members": [
+      {
+        "//": "This is called 'a' just to defeat project sorting",
```
hawkw, line 6: beautiful

PR #185, opened 2023-07-24
fix(spitebuf + kchannel): Fixes sizing requirements of spitebuf

jamesmunns:
This PR fixes some of the invariants required by the upstream heapless::mpmc implementation that I didn't originally uphold.

SpiteBuf will now panic on an incorrect size of 1, and KChannel will automatically increase the buffer size and emit a warning.

### DIFF
--- a/source/kernel/src/comms/kchannel.rs
+++ b/source/kernel/src/comms/kchannel.rs
@@ -55,11 +55,34 @@ impl<T> Deref for KChannel<T> {
     }
 }
 
+fn right_size(mut cap: usize) -> usize {
+    if cap < 2 {
+        tracing::warn!(
+            was = cap,
+            now = 2,
+            "Increasing KChannel size to minimum size of two!"
+        );
+    }
+    let npot = cap.next_power_of_two();
+    if npot != cap {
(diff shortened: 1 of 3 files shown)


source/kernel/src/comms/kchannel.rs
```diff
@@ -55,11 +55,34 @@ impl<T> Deref for KChannel<T> {
     }
 }
 
+fn right_size(mut cap: usize) -> usize {
+    if cap < 2 {
+        tracing::warn!(
```
hawkw, line 61: CI fails because we need a `use crate::tracing;`: https://github.com/tosc-rs/mnemos/actions/runs/5650604492/job/15307247587?pr=185#step:8:477

hawkw, line 61: take it or leave it: what do you think about having this panic in debug mode, or at least with `cfg(test)`? that way, when we're running tests, it fails loudly, but if it somehow slips into a real build, we just fix it.

**Edit**: oh i see that we assert it elsewhere. that's fine!

hawkw (approved):

PR #188, opened 2023-07-25
style: fix all clippy lints and add clippy to CI

hawkw:
Note that some of these lints required non-trivial refactoring to fix. @jamesmunns, wdyt?

### DIFF
--- a/source/kernel/src/registry/mod.rs
+++ b/source/kernel/src/registry/mod.rs
@@ -241,6 +241,12 @@ pub enum OneshotRequestError {
     Receive(ReusableError),
 }
 
+#[derive(Debug, Eq, PartialEq)]
+pub enum SendError {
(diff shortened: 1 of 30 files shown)


source/kernel/src/registry/mod.rs
```diff
@@ -241,6 +241,12 @@ pub enum OneshotRequestError {
     Receive(ReusableError),
 }
 
+#[derive(Debug, Eq, PartialEq)]
+pub enum SendError {
+    /// The service on the other end of the [`KHandle`] has terminated!
```
hawkw, line 246:
```suggestion
    /// The service on the other end of the [`KernelHandle`] has terminated!
```

jamesmunns (approved):
These all look great, I better go configure my editor :)

Issue #139, opened 2023-07-10
platform(d1): persistent boot


hawkw:
Currently, we can only boot MnemOS on the D1 by writing the kernel directly into the D1's DDR using `xfel`. This is not ideal. Instead, we should be able to put the kernel image either in flash or on the SD card, and have the device bootload itself when it comes out of reset.

per @jamesmunns in issue #120:
> This is probably going to involve a couple things:
>
> * Figuring out how to prepare a binary image that the D1 will recognize and start booting from
> * Writing some kind of early stage bootloader (could be the whole kernel at first lol)
> * Figuring out how much of the SD card the hardware will load for us
> * Initializing the DDR3 memory
> * Booting without the help of XFEL
>
> I think the oreboot/rustsbi crates have at least figured out part of this (preparing an image, and initializing the DDR3 memory), though it may have been from a SPI flash and not an SD card, though the stock MQ-Pro boards do not come with SPI flash, so we would have to solder our own on.

per @orangecms in https://github.com/tosc-rs/mnemos/issues/120#issuecomment-1616595458:
> All that the D1 needs to boot from any media is the special eGON header. On SD cards, it needs to be at a specific offset. See here further down: https://linux-sunxi.org/Allwinner_Nezha#U-Boot
>
> We have the header implemented in oreboot; see the sunxi/nezha mainboard directory, ant bt0/main.rs under it. A checksum is filled in via xtask.

So, it *sounds* like we can boot directly from SD if we can make an eGON header and put it at the right offset. Adding SPI flash to the MnemOS breakout board is (probably?) not necessary?  The sunxi wiki explains the eGON header format here: https://linux-sunxi.org/EGON

(sidenote: every time I see "eGON header" i immediately think of Egon from _Ghostbusters_)

hawkw:
Oreboot's D1 stage0 lives [here](https://github.com/oreboot/oreboot/blob/2ce5277b86d07b1e1c680397a38cbaa59ce43a6a/src/mainboard/sunxi/nezha/bt0/src/main.rs). In particular, we can see their eGON header here: https://github.com/oreboot/oreboot/blob/2ce5277b86d07b1e1c680397a38cbaa59ce43a6a/src/mainboard/sunxi/nezha/bt0/src/main.rs#L65-L80
They have a `cargo xtask` that does most of the flashing work: https://github.com/oreboot/oreboot/blob/main/xtask/src/sunxi/nezha.rs

hawkw:
[...]

PR #192, opened 2023-07-28
feat(d1): add clint and ccu module

jspngh:
This PR adds an initial implementation of 2 new modules:
- a `ccu` module for clock initialization and module clock gating/reset
- a `clint` module to provide access to `mtime` and `mtimecmp`

By doing the same clock initialization that `xfel` would do for us,
it becomes possible to boot from a persistent medium (e.g., an SD card).
Relates to #139.

### DIFF
--- a/platforms/allwinner-d1/core/src/ccu.rs
+++ b/platforms/allwinner-d1/core/src/ccu.rs
@@ -1,0 +1,199 @@
+//! This module provides a higher-level interface for the `Clock Controller Unit`.
+use d1_pac::CCU;
+
+pub struct Ccu {
+    ccu: CCU,
+}
+
+macro_rules! set_module {
+    ($self: ident, $module: ident) => {
+        if ($self.ccu.$module.read().pll_en().bit_is_clear()) {
+            $self.ccu.$module.modify(|_, w| {
+                w.pll_ldo_en().enable();
+                w.pll_en().enable();
+                w
+            });
+
+            $self.ccu.$module.modify(|_, w| w.lock_enable().enable());
+
+            while $self.ccu.$module.read().lock().bit_is_clear() {
+                core::hint::spin_loop();
+            }
+            sdelay(20);
+
+            $self.ccu.$module.modify(|_, w| w.lock_enable().disable());
+        }
+    };
+}
+
+// temporary
+fn sdelay(delay_us: usize) {
+    let clint = unsafe { crate::clint::Clint::summon() };
+    let t = clint.get_mtime();
+    while clint.get_mtime() < (t + 24 * delay_us) {
+        core::hint::spin_loop()
+    }
+}
+
+impl Ccu {
+    pub fn new(ccu: CCU) -> Self {
+        Self { ccu }
+    }
+
+    pub fn release(self) -> CCU {
+        self.ccu
+    }
+
+    /// This function is currently a copy from
+    /// [xboot](https://github.com/xboot/xboot/blob/master/src/arch/riscv64/mach-d1/sys-clock.c)
+    pub fn sys_clock_init(&mut self) {
+        self.set_pll_cpux_axi();
+        self.set_pll_periph0();
+        self.set_ahb();
+        self.set_apb();
+        self.set_dma();
+        self.set_mbus();
+        set_module!(self, pll_peri_ctrl);
+        set_module!(self, pll_video0_ctrl);
+        set_module!(self, pll_video1_ctrl);
+        set_module!(self, pll_ve_ctrl);
+        set_module!(self, pll_audio0_ctrl);
+        set_module!(self, pll_audio1_ctrl);
+    }
+
+    fn set_pll_cpux_axi(&mut self) {
+        /* Select cpux clock src to osc24m, axi divide ratio is 3, system apb clk ratio is 4 */
+        self.ccu.riscv_clk.write(|w| {
+            w.clk_src_sel().hosc();
+            w.axi_div_cfg().variant(3);
+            w.div_cfg().variant(1);
+            w
+        });
+        sdelay(1);
+
+        /* Disable pll gating */
+        self.ccu
+            .pll_cpu_ctrl
+            .modify(|_, w| w.pll_output_gate().disable());
+
+        /* Enable pll ldo */
+        self.ccu.pll_cpu_ctrl.modify(|_, w| w.pll_ldo_en().enable());
+        sdelay(5);
+
+        /* Set default clk to 1008mhz */
+        self.ccu.pll_cpu_ctrl.modify(|r, w| {
+            // undocumented part of register is cleared by xboot
+            unsafe { w.bits(r.bits() & !(0x3 << 16)) };
+            w.pll_m().variant(0);
+            w.pll_n().variant(41);
+            w
+        });
+
+        /* Lock enable */
+        self.ccu
+            .pll_cpu_ctrl
+            .modify(|_, w| w.lock_enable().enable());
+
+        /* Enable pll */
+        self.ccu.pll_cpu_ctrl.modify(|_, w| w.pll_en().enable());
+
+        /* Wait pll stable */
+        while self.ccu.pll_cpu_ctrl.read().lock().bit_is_clear() {
+            core::hint::spin_loop();
+        }
+        sdelay(20);
+
+        /* Enable pll gating */
+        self.ccu
+            .pll_cpu_ctrl
+            .modify(|_, w| w.pll_output_gate().enable());
+
+        /* Lock disable */
+        self.ccu
+            .pll_cpu_ctrl
+            .modify(|_, w| w.lock_enable().disable());
+        sdelay(1);
+
+        /* Set and change cpu clk src */
+        self.ccu.riscv_clk.modify(|_, w| {
+            w.clk_src_sel().pll_cpu();
+            w.axi_div_cfg().variant(1);
+            w.div_cfg().variant(0);
+            w
+        });
+        sdelay(1);
+    }
+
+    fn set_pll_periph0(&mut self) {
+        /* Periph0 has been enabled */
+        if self.ccu.pll_peri_ctrl.read().pll_en().bit_is_set() {
+            return;
+        }
+
+        /* Change psi src to osc24m */
+        self.ccu.psi_clk.modify(|_, w| w.clk_src_sel().hosc());
+
+        /* Set default val */
+        self.ccu.pll_peri_ctrl.write(|w| w.pll_n().variant(0x63));
+
+        /* Lock enable */
+        self.ccu
+            .pll_peri_ctrl
+            .modify(|_, w| w.lock_enable().enable());
+
+        /* Enabe pll 600m(1x) 1200m(2x) */
+        self.ccu.pll_peri_ctrl.modify(|_, w| w.pll_en().enable());
+
+        /* Wait pll stable */
+        while self.ccu.pll_peri_ctrl.read().lock().bit_is_clear() {
+            core::hint::spin_loop();
+        }
+        sdelay(20);
+
+        /* Lock disable */
+        self.ccu
+            .pll_peri_ctrl
+            .modify(|_, w| w.lock_enable().disable());
+    }
+
+    fn set_ahb(&mut self) {
+        self.ccu
+            .psi_clk
+            .write(|w| w.factor_m().variant(2).factor_n().n1());
+        self.ccu
+            .psi_clk
+            .modify(|_, w| w.clk_src_sel().pll_peri_1x());
+        sdelay(1);
+    }
+
+    fn set_apb(&mut self) {
+        self.ccu.apb_clk[0].write(|w| w.factor_m().variant(2).factor_n().n2());
+        self.ccu.apb_clk[0].modify(|_, w| w.clk_src_sel().pll_peri_1x());
+        sdelay(1);
+    }
+
+    fn set_dma(&mut self) {
+        /* Dma reset */
+        self.ccu.dma_bgr.modify(|_, w| w.rst().deassert());
+        sdelay(20);
+        /* Enable gating clock for dma */
+        self.ccu.dma_bgr.modify(|_, w| w.gating().pass());
+    }
+
+    fn set_mbus(&mut self) {
+        /* Reset mbus domain */
+        self.ccu.mbus_clk.modify(|_, w| w.mbus_rst().deassert());
+        sdelay(1);
+        /* Enable mbus master clock gating */
+        self.ccu.mbus_mat_clk_gating.write(|w| {
+            w.dma_mclk_en().pass();
+            w.ve_mclk_en().pass();
+            w.ce_mclk_en().pass();
+            w.tvin_mclk_en().pass();
+            w.csi_mclk_en().pass();
+            w.g2d_mclk_en().pass();
+            w.riscv_mclk_en().pass();
+            w
+        });
+    }
+}
--- a/platforms/allwinner-d1/core/src/ccu.rs
+++ b/platforms/allwinner-d1/core/src/ccu.rs
@@ -1,8 +1,21 @@
 //! This module provides a higher-level interface for the `Clock Controller Unit`.
 use d1_pac::CCU;
+use d1_pac::DMAC;
+use d1_pac::{SMHC0, SMHC1, SMHC2};
+use d1_pac::{SPI0, SPI_DBI};
+use d1_pac::{TWI0, TWI1, TWI2, TWI3};
+use d1_pac::{UART0, UART1, UART2, UART3, UART4, UART5};
 
 pub struct Ccu {
     ccu: CCU,
+}
+
+/// Trait to be implemented for module clocks that can be gated and reset
+pub trait BusGatingResetRegister {
+    /// Enable or disable the clock reset bit
+    fn gating(ccu: &mut CCU, pass: bool);
+    /// Enable or disable the clock gating bit
+    fn reset(ccu: &mut CCU, deassert: bool);
 }
 
 macro_rules! set_module {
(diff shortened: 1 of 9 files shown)


hawkw:
This is great, I think the addition of a higher level abstraction that can be used by our existing drivers is really nice! Will give a more in-depth review shortly.

platforms/allwinner-d1/core/src/clint.rs
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    /// Get the (machine) time value.
+    ///
+    /// Note that the CLINT of the C906 core does not implement
+    /// the `mtime` register and we need to get the time value
+    /// with a CSR, which the `riscv` crate implements for us.
+    pub fn get_mtime(&self) -> usize {
+        riscv::register::time::read()
```
hawkw, line 31: nit, take it or leave it: this comment feels like it's describing the implementation of this method more than its behavior, so i might move it into the function body:
```suggestion
    pub fn get_mtime(&self) -> usize {
        // Note that the CLINT of the C906 core does not implement
        // the `mtime` register and we need to get the time value
        // with a CSR, which the `riscv` crate implements for us.
        riscv::register::time::read()
```

platforms/allwinner-d1/core/src/clint.rs
```diff
@@ -0,0 +1,54 @@
...
+        let cmph = (cmp >> 32) as u32;
+        let cmpl = (cmp & 0xffff_ffff) as u32;
+        unsafe {
+            self.clint.mtimecmph.write(|w| w.bits(cmph));
+            self.clint.mtimecmpl.write(|w| w.bits(cmpl));
+        }
+    }
+
+    /// Reset the machine time comparator to its default value.
+    pub fn reset_mtimecmp(&self) {
```
hawkw, line 48: since these methods modify two registers as part of one operation, i wonder if they should maybe take `&mut self` instead of `&self`? this would prevent them from being called concurrently, which could potentially result in a torn write. on the other hand, since the D1 is a single-core CPU, this probably doesn't matter... :woman_shrugging:

platforms/allwinner-d1/core/src/ccu.rs
```diff
@@ -0,0 +1,260 @@
...
+            .modify(|_, w| w.lock_enable().disable());
+    }
+
+    fn set_ahb(&mut self) {
+        self.ccu
+            .psi_clk
+            .write(|w| w.factor_m().variant(2).factor_n().n1());
+        self.ccu
+            .psi_clk
+            .modify(|_, w| w.clk_src_sel().pll_peri_1x());
```
hawkw, line 199: naively, looking at this makes me think that the write and modify calls could be combined into one `write` operation...but, looking at the `xboot` code, i see that they perform one write and then a second write that sets only the CLK_SRC_SEL field: https://github.com/xboot/xboot/blob/6d6b93947b7fcb8c3924fedb0715c23877eedd5e/src/arch/riscv64/mach-d1/sys-clock.c#L144-L145
so i imagine that it's actually important to perform two separate writes. maybe worth a comment saying that we do this across a pair of writes because that's what `xboot` does?

platforms/allwinner-d1/core/src/ccu.rs
```diff
@@ -0,0 +1,260 @@
...
+        sdelay(1);
+    }
+
+    fn set_apb(&mut self) {
+        self.ccu.apb_clk[0].write(|w| w.factor_m().variant(2).factor_n().n2());
+        self.ccu.apb_clk[0].modify(|_, w| w.clk_src_sel().pll_peri_1x());
```
hawkw, line 205: similarly, it might be nice to have a comment saying that we do this in a pair of separate writes to the register because xboot does that.

platforms/allwinner-d1/core/src/ccu.rs
```diff
@@ -0,0 +1,260 @@
+//! This module provides a higher-level interface for the `Clock Controller Unit`.
```
hawkw, line 1: tiny style nit:
```suggestion
//! This module provides a higher-level interface for the Clock Controller Unit (CCU).
```

platforms/allwinner-d1/core/src/ccu.rs
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    /// Initialize the system clocks to the same default value that is also set by `xfel`
+    pub fn sys_clock_init(&mut self) {
+        // The clock initialization functions are ported to Rust, based on the C implementation in
+        // [xboot](https://github.com/xboot/xboot/blob/master/src/arch/riscv64/mach-d1/sys-clock.c)
+        self.set_pll_cpux_axi();
+        self.set_pll_periph0();
+        self.set_ahb();
+        self.set_apb();
```
hawkw, line 88: looking at the `xboot` source, i see that there's also a `set_dma` function, which is [also called in `sys_clock_init`](https://github.com/xboot/xboot/blob/6d6b93947b7fcb8c3924fedb0715c23877eedd5e/src/arch/riscv64/mach-d1/sys-clock.c#L209), and [resets + enables the DMA BGR](https://github.com/xboot/xboot/blob/6d6b93947b7fcb8c3924fedb0715c23877eedd5e/src/arch/riscv64/mach-d1/sys-clock.c#L156-L163). 

it looks like the reason we don't do that here is because the DMAC driver calls `enable()` on that register, here: https://github.com/tosc-rs/mnemos/blob/94199eb8946fc054aeb78a944818d7b9c73eeec5/platforms/allwinner-d1/core/src/dmac/mod.rs#L30

it might be worth including a comment here like
```suggestion
        self.set_apb();
        // This is where `xboot` resets and enables the DMA bus gating register.
        // We don't do that, because this is performed by the DMAC driver's 
        // initialization function, instead.
```
it might also be worth checking that we're actually setting the same bits they set when we call `enable<DMAC>()`?

hawkw (approved):
this looks fantastic, i like the addition of a higher-level abstraction for enabling and disabling clock gating registers that can be used in our other drivers. 

i left some minor style and documentation suggestions, but none of them are particularly important to me --- i'll be happy to merge this PR regardless. thanks for all your hard work on getting this working! :D

jspngh:
> this looks fantastic, i like the addition of a higher-level abstraction for enabling and disabling clock gating registers that can be used in our other drivers.
> 
> i left some minor style and documentation suggestions, but none of them are particularly important to me --- i'll be happy to merge this PR regardless. thanks for all your hard work on getting this working! :D

Thanks for the thorough review! All sensible suggestions, I'm going to apply most, if not all of them.

platforms/allwinner-d1/core/src/ccu.rs
```diff
@@ -0,0 +1,260 @@
...
+        self.set_mbus();
+        set_module!(self, pll_peri_ctrl);
+        set_module!(self, pll_video0_ctrl);
+        set_module!(self, pll_video1_ctrl);
+        set_module!(self, pll_ve_ctrl);
+        set_module!(self, pll_audio0_ctrl);
+        set_module!(self, pll_audio1_ctrl);
```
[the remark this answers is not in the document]
jspngh: I didn't know you could do that! TIL

platforms/allwinner-d1/core/src/ccu.rs
```diff
@@ -0,0 +1,260 @@
...
+            .modify(|_, w| w.pll_output_gate().disable());
+
+        /* Enable pll ldo */
+        self.ccu.pll_cpu_ctrl.modify(|_, w| w.pll_ldo_en().enable());
+        sdelay(5);
+
+        /* Set default clk to 1008mhz */
+        self.ccu.pll_cpu_ctrl.modify(|r, w| {
+            // undocumented part of register is cleared by xboot
+            unsafe { w.bits(r.bits() & !(0x3 << 16)) };
```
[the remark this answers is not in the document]
jspngh: We could probably just drop this and check if everything keeps working, but I wanted to do a literal translation first.

platforms/allwinner-d1/core/src/ccu.rs
```diff
@@ -0,0 +1,260 @@
...
+            .pll_cpu_ctrl
+            .modify(|_, w| w.pll_output_gate().enable());
+
+        /* Lock disable */
+        self.ccu
+            .pll_cpu_ctrl
+            .modify(|_, w| w.lock_enable().disable());
+        sdelay(1);
+
+        /* Set and change cpu clk src */
```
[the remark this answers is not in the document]
jspngh: Yep, that was me being lazy, I'll clean them up according to your suggestions.

platforms/allwinner-d1/core/src/ccu.rs
jspngh: Indeed, I removed it here since we enable the peripheral clocks in the peripheral modules themselves.
The bits in `enable<DMAC>()` match those from `set_dma`.

platforms/allwinner-d1/core/src/ccu.rs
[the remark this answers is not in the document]
hawkw: yup! macro declarations can be scoped --- and i *believe* it should be accessible only within the function, IIRC.

platforms/allwinner-d1/core/src/ccu.rs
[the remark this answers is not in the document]
hawkw: i think it's wonderful! but, i would be curious to find out what, if anything, breaks without it

jamesmunns:
@jspngh as portions of this code are a fairly literal port of the `xboot` code (which is MIT licensed), I think it would be fair to include a licensing note at the top of the relevant ported files.

Since mnemos is already MIT+Apache2, maybe something like:

```rust
//! ## Provenance
//!
//! Portions of this file were ported from the `xboot` project. These portions were used
//! under the terms of the MIT license. Refer to `LICENSE-MIT` at top of this repo for
//! full license text. `xboot` copyright information:
//!
//! Copyright(c) 2007-2022 Jianjun Jiang <8192542@qq.com>
```

In the future, we can add this line to our `LICENSE-MIT`.

platforms/allwinner-d1/core/src/ccu.rs
```diff
@@ -0,0 +1,272 @@
...
+    fn gating(ccu: &mut CCU, pass: bool);
+    /// Enable or disable the clock gating bit
+    fn reset(ccu: &mut CCU, deassert: bool);
+}
+
+// TODO: should this move into the `Clint`?
+fn sdelay(delay_us: usize) {
+    let clint = unsafe { crate::clint::Clint::summon() };
+    let t = clint.get_mtime();
+    // TODO: confirm clock source of mtime
```
jamesmunns, line 25: mtime is sourced from the CPU core clock. It will tick at the same rate of the CPU core. This means it will also be affected by the clock selected for the CPU.

We could rename this `rename_jiffies`, if we want to adopt the kernel naming style.

platforms/allwinner-d1/core/src/ccu.rs
```diff
@@ -0,0 +1,272 @@
...
+        Self { ccu }
+    }
+
+    #[must_use]
+    pub fn release(self) -> CCU {
+        self.ccu
+    }
+
+    /// De-assert the reset bit and enable the clock gating bit for the given module
+    pub fn enable_module<MODULE: BusGatingResetRegister>(&mut self) {
```
jamesmunns, line 43: Misc note, you could enforce that the caller "owns" the peripheral by doing this instead (same for disable), which would also mean we don't have to turbofish the call either:

```suggestion
    pub fn enable_module<MODULE: BusGatingResetRegister>(&mut self, _mod: &mut MODULE) {
```

platforms/allwinner-d1/core/src/drivers/uart.rs
```diff
@@ -208,10 +209,9 @@ impl D1Uart {
 ///
 /// - The `UART0` register block must not be concurrently written to.
 /// - This function should be called only while running on an Allwinner D1.
-pub unsafe fn kernel_uart(ccu: &mut CCU, gpio: &mut GPIO, uart0: UART0) -> Uart {
+pub unsafe fn kernel_uart(ccu: &mut Ccu, gpio: &mut GPIO, uart0: UART0) -> Uart {
     // Enable UART0 clock.
-    ccu.uart_bgr
-        .write(|w| w.uart0_gating().pass().uart0_rst().deassert());
+    ccu.enable_module::<UART0>();
```
jamesmunns, line 214: If you take my other suggestion, this would become:

```rust
ccu.enable_module(&mut UART0);
```

platforms/allwinner-d1/core/src/drivers/spim.rs
```diff
@@ -27,9 +28,9 @@ pub struct Spim1 {
 ///
 /// - The `SPI_DBI``s register block must not be concurrently written to.
 /// - This function should be called only while running on an Allwinner D1.
-pub unsafe fn kernel_spim1(spi1: SPI_DBI, ccu: &mut CCU, gpio: &mut GPIO) -> Spim1 {
+pub unsafe fn kernel_spim1(spi1: SPI_DBI, ccu: &mut Ccu, gpio: &mut GPIO) -> Spim1 {
     // Set clock rate (fixed to 2MHz), and enable the SPI peripheral
-    ccu.spi1_clk.write(|w| {
+    ccu.borrow_raw().spi1_clk.write(|w| {
```
jamesmunns, line 33: Might want to add a TODO here to link to the "we shouldn't need `borrow_raw()` in the long term"

jamesmunns (approved):
Looks good to me, thank you so much @jspngh!

platforms/allwinner-d1/core/src/ccu.rs
jspngh: Should we require a mutable reference to be sure of ownership?
It's just a bit more inconvenient, because then we have to change the function parameters to be as `mut` as well, e.g.:

```diff
- pub unsafe fn kernel_uart(ccu: &mut Ccu, gpio: &mut GPIO, uart0: UART0) -> Uart {
+ pub unsafe fn kernel_uart(ccu: &mut Ccu, gpio: &mut GPIO, mut uart0: UART0) -> Uart {
```

platforms/allwinner-d1/core/src/ccu.rs
jspngh: Did you mean something like `delay_jiffies`?

platforms/allwinner-d1/core/src/ccu.rs
jamesmunns: Up to you! Just providing context if that helps, if it means we need to summon peripherals everywhere, then I agree it might be more trouble than it is worth.

platforms/allwinner-d1/core/src/ccu.rs
jamesmunns: Ah yes, sorry :)

platforms/allwinner-d1/core/src/ccu.rs
jspngh: I think it's a good idea to pass in the module to enforce ownership, but I don't know if the reference needs to be mutable.
We don't actually modify any registers from the module itself, so from that point of view I think
`_mod: &MODULE` is enough, but maybe that does not really *enforce* the ownership?

platforms/allwinner-d1/core/src/ccu.rs
jspngh: > mtime is sourced from the CPU core clock. It will tick at the same rate of the CPU core. This means it will also be affected by the clock selected for the CPU.

Btw, did you find this documented somewhere? No matter what search terms I'm using, 
I can't find any useful info about how the C906 CLINT increments mtime :sweat_smile:

platforms/allwinner-d1/core/src/ccu.rs
jamesmunns: IF you did not use unsafe anywhere else to create a second instance of a peripheral, then it does enforce ownership. A mutable reference can only be created once, from something owned, and the PAC model will only ever (safely) give you ownership of the peripheral once.

platforms/allwinner-d1/core/src/ccu.rs
jamesmunns: So, I'm not actually sure if mtime's clint is different than "the riscv mtime register", and it seems I was actually also wrong here 🤦:

https://five-embeddev.com/riscv-isa-manual/latest/machine.html#machine-timer-registers-mtime-and-mtimecmp

> Platforms provide a real-time counter, exposed as a memory-mapped machine-mode read-write register, mtime. mtime must increment at constant frequency, and the platform must provide a mechanism for determining the period of an mtime tick. The  mtime register will wrap around if the count overflows.

platforms/allwinner-d1/core/src/ccu.rs
jspngh: Yup, I understand that. I'm going to rephrase, because I don't think it was clear what my *actual* question was.
I think there are two options:
```rust
pub unsafe fn kernel_uart(ccu: &mut Ccu, gpio: &mut GPIO, mut uart0: UART0) -> Uart {
    // Enable UART0 clock.
    ccu.enable_module(&mut uart0);
    ...
}
```
and
```rust
pub unsafe fn kernel_uart(ccu: &mut Ccu, gpio: &mut GPIO, uart0: UART0) -> Uart {
    // Enable UART0 clock.
    ccu.enable_module(&uart0);
    ...
}
```
The downside of option 1 is the additional `mut` in the parameter list, to be able to take a mutable reference.
But then you are sure that the caller has a unique reference to the module.

Then my question was, does option 1 prove/enforce ownership better than option 2?
The caller could also be passing a mutable reference that it received from its caller, so then it also doesn't own the module.
In terms of safety, I don't think there is a difference between both options, since you have the `&mut Ccu`, which is the register block that is actually being modified.

jspngh:
If there are no more remarks and everyone is okay with the PR name + description as squash commit message,
then it should be ready for the merge :slightly_smiling_face:

hawkw:
> If there are no more remarks and everyone is okay with the PR name + description as squash commit message, then it should be ready for the merge 🙂

Fantastic, let's go! Thanks again for getting this stuff working!

PR #193, opened 2023-07-28
feat(esp32-c3): add stub esp32-c3 project

hawkw:
This branch adds a mostly empty ESP32-C3 project in `platforms/`, and
adds Just recipes for building and flashing the ESP32-C3. Currently, the
C3 platform crate just prints "hello world"; it does *not* actually run
a working MnemOS kernel. Actual MnemOS bringup on top of the stub
project will be added in a follow-up PR.


platforms/esp32c3-buddy/LICENSE-APACHE
jamesmunns, line 1: IMO we could drop this since it is the same as the top level repo license.

platforms/esp32c3-buddy/LICENSE-MIT
jamesmunns, line 1: Same as LICENSE-APACHE

platforms/esp32c3-buddy/src/main.rs
```diff
@@ -0,0 +1,37 @@
...
+        peripherals.TIMG1,
+        &clocks,
+        &mut system.peripheral_clock_control,
+    );
+    let mut wdt1 = timer_group1.wdt;
+    rtc.swd.disable();
+    rtc.rwdt.disable();
+    wdt0.disable();
+    wdt1.disable();
+    println!("Hello world!");
```
jamesmunns, line 34: 👋

source/kernel/src/gpio.rs
jamesmunns, line 1: Remove the empty file if it's unused?

platforms/esp32c3-buddy/.vscode/settings.json
jamesmunns, line 1: Maybe pull this file too?

jamesmunns (approved):
Overall lgtm! I think there's a couple of templated files we might want to remove, but otherwise fine to merge.

source/kernel/src/gpio.rs
hawkw: gah, whoops!

platforms/esp32c3-buddy/LICENSE-APACHE
hawkw: fair, i'll remove them --- IIRC some of the other crates have their own license
files, but i guess that's probably an artifact of having copied them over from their
own repos?

PR #194, opened 2023-07-28
fix: remove use of `tracing` v0.2

### LINKED ISSUE #142: `cargo melpo` doesn't compile when in repo root, due to multiple `tracing` versions
After checking out the repo and running `cargo melpo`, I get hundreds of errors similar to this:

```
error[E0659]: `Instrument` is ambiguous
   --> source/kernel/src/services/forth_spawnulator.rs:156:5
    |
156 |     #[tracing::instrument(skip(kernel, vms))]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ambiguous name
    |
    = note: ambiguous because of multiple glob imports of a name in the same module
```

The issue seems to be that both `tracing@0.2.0` and `tracing@0.1.37` are included in the dependency graph.

When I run the command in the `./sources/melpomene` directory, it works.

hawkw:
Currently, we use `tracing` v0.2 on some platforms, and v0.1 on others. This was motivated by `tracing` v0.1 requiring a single allocation performed using `liballoc` to allocate an `Arc` when setting the default subscriber, which is not needed in `tracing` v0.2. This was an issue when mnemOS did not use `liballoc`. However, we now use `liballoc` on all platforms, so we can completely cut over from the unreleased `tracing` v0.2 to released `tracing` v0.1 everywhere.

This branch makes that change, rendering the complex feature flagging situation previously used to switch between the released and unreleased `tracing` versions unnecessary. This should fix the build in the root workspace without feature flags.

We still feature flag the automatic spawning of the serial port subscriber, so that Melpomene and Pomelo can use their own native subscribers instead. We may want to change this so that the platform impl is responsible for setting the global default subscriber, instead, but this works for now.

Fixes #142

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -463,7 +463,7 @@ dependencies = [
  "tonic",
  "tracing 0.1.37",
  "tracing-core 0.1.31",
- "tracing-subscriber 0.3.17",
+ "tracing-subscriber",
 ]
 
 [[package]]
@@ -582,9 +582,9 @@ dependencies = [
  "serialport 4.0.1",
  "serialport 4.2.1",
  "sermux-proto",
- "tracing 0.2.0",
+ "tracing 0.1.37",
  "tracing-serde-structured",
- "tracing-subscriber 0.3.0",
+ "tracing-subscriber",
 ]
 
 [[package]]
@@ -1366,7 +1366,7 @@ dependencies = [
  "generator",
  "scoped-tls",
  "tracing 0.1.37",
- "tracing-subscriber 0.3.17",
+ "tracing-subscriber",
 ]
 
 [[package]]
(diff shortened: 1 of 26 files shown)


hawkw:
In theory this would also let us move a lot more stuff back into the root workspace, but we may not want to do that due to build artifact caching for different targets...

jamesmunns (approved):
LGTM, thank you @hawkw! Any other potential/theoretical side effects of this switchover we should be aware of?

hawkw:
> LGTM, thank you @hawkw! Any other potential/theoretical side effects of this switchover we should be aware of?

There shouldn't really be any significant impacts besides letting us use the releases versions of `tracing` libraries!

PR #196, opened 2023-07-29
RFC: WiFi Buddy Interface Design

hawkw:
This commit starts a new RFC for designing the hardware and software interfaces between MnemOS on the CPU and MnemOS on the WiFi Buddy. I've started by adding a summary of the background and a list of design questions that need to be answered as we determine how the WiFi Buddy and the CPU will communicate. This is intended as a jumping-off point for future conversations, and (eventually) to document the design.

@jamesmunns, I'd love your input on this --- feel free to add to the RFC doc directly, or leave comments on the PR, whatever you prefer!


rfcs/0000-wifi-buddy-interface.md
```diff
@@ -0,0 +1,125 @@
...
+The WiFi Buddy hardware device is a [Seeedstudio XIAO] or [Adafruit
+QT Py] ESP32-C3 development board. These are development boards for the ESP32-C3
+which are pinout compatible, so they can be used interchangeably. The ESP32-C3
+board will be treated as a system-on-module (SoM) and integrated with the MnemOS
+Beepy shield using pin headers (or potentially soldered onto the board).
+
+The following pins are broken out by the WiFi Buddy hardware and can be used for
+communication between the CPU and WiFi Buddy:
+
+- One SPI interface (`MOSI`, `MISO`, and `SCK`)
```
jamesmunns, line 27: There's also a chip select pin `CSn`.

rfcs/0000-wifi-buddy-interface.md
```diff
@@ -0,0 +1,125 @@
...
+where TCP or any other higher-level network protocols are implemented on the
+CPU, and the CPU and WiFi Buddy communicate by exchanging MAC-layer [802.3]
+Ethernet frames.
+
+## Design Questions
+
+Based on the background information discussed above, we can begin to enumerate
+design questions for the WiFi Buddy interface:
+
+- **Is control path communication in-band or out of band?** If data path frames
```
jamesmunns, line 69: Mentioned in chat, but I'd probably recommend just using SPI unless we come up with a good reason not to.

rfcs/0000-wifi-buddy-interface.md
```diff
@@ -0,0 +1,125 @@
...
+  are exchanged between the host CPU and the WiFi Buddy over the SPI bus, do we
+  also exchange control messages on the same bus? Or, are control messages sent
+  using another interface, such as I²C?
+
+  Using a single interface for both control and data path communication is
+  conceptually simpler, and has the advantage of a clear and obvious ordering
+  between data and control messages. It also means that we don't necessarily
+  need to implement drivers for all the interfaces available on the ESP32-C3.
+
+  Downsides of in-band control messages include that we must introduce some
```
jamesmunns, line 79: Mention in chat: I think using postcard as the wire format means that decoding/encoding is likely to be low overhead (CPU and bandwidth use wise), and allow for flexibility in the future if we add more message types.

rfcs/0000-wifi-buddy-interface.md
```diff
@@ -0,0 +1,125 @@
...
+  additional form of tagging of data sent on the SPI bus, in order  to indicate
+  what bytes are part of a control message and what bytes are actual network
+  data. This may introduce some additional overhead. On the other hand, if we
+  want to eventually include Bluetooth Low Energy as well as WiFi, we will need
+  the ability to indicate what data is an Ethernet frame and what is BLE,
+  anyway, so we'll already be introducing some overhead for tagging data.
+
+- **Do we want the CPU to be able to receive `trace-proto` from WiFi Buddy?**
```
jamesmunns, line 87: I think we will want this, but it can also be deferred during early development since we'll have the additional USB link

rfcs/0000-wifi-buddy-interface.md
hawkw: I didn't include chip select here because any of the 4 GPIOs could be the chip select pin, right?

rfcs/0000-wifi-buddy-interface.md
```diff
@@ -0,0 +1,125 @@
...
+  [`mnemos-trace-proto`]. Do we want `trace-proto`-formatted logs from the WiFi
+  Buddy to be exposed to the CPU, so that it can proxy them to an attached debug
+  host? Or, is using the WiFi Buddy SoM's onboard USB-serial hardware sufficient
+  for debugging purposes?
+
+- **What control messages are necessary?** Eventually, we'll need to enumerate
+  what control messages will need to be exchanged in order to implement the
+  required functionality.
+
+- **How does the WiFi Buddy signal that data is ready?** Since the CPU is the
```
jamesmunns, line 98: Definitely +1 that we need at least ONE GPIO signal line, even if it is optional. Possible useful lines are:

* "Modem ready to receive data" (can also indicate "I am busy processing" or "buffers are full")
* "Modem has data for host"

rfcs/0000-wifi-buddy-interface.md
```diff
@@ -0,0 +1,125 @@
...
+firmware, which we are referring to as "WiFi Buddy". See [#121] for details on
+the rationale for choosing the ESP32-C3 as the WiFi device, and for discussion
+of the overall design for ESP32-C3 integration.
+
+The goal of this document is to propose an overall design for the hardware and
+software interfaces between MnemOS on the CPU and MnemOS on the WiFi Buddy.
+
+## Background
+
+### Hardware
```
jamesmunns, line 16: Might be worth mentioning/discussing https://github.com/tosc-rs/mnemos/tree/main/hardware/esp32c3-wifi-dev, and particularly note that we'll be able to use:

* 4-pin SPI
* 2-pin UART
* 2 signal GPIOs (could be 4, if we don't use UART)

For initial development.

jamesmunns (approved):
Overall looks great to me, and I appreciate you writing this up @hawkw!

rfcs/0000-wifi-buddy-interface.md
hawkw: Yeah, we're aligned on this --- I wanted to enumerate both options but the proposed design will be SPI only (and IRQs)

rfcs/0000-wifi-buddy-interface.md
jamesmunns: *sort of*. Short answer, yes it could be for now, but might be worth explicitly addressing as such.

rfcs/0000-wifi-buddy-interface.md
jamesmunns, line 1: Don't forget to change the number of the file from `0000` to `0196` before merging

PR #199, opened 2023-07-30
RFC: Basic Configuration

### LINKED ISSUE #160: Idea collector: Configuration System

jamesmunns:
This RFC introduces a high level plan for implementing a configuration system, as discussed in #160.

Closes #160


jamesmunns:
note to self: add a "tomorrows problem" section about layering configuration, and potentially refernce https://crates.io/crates/config as a source of inspiration.

jamesmunns:
note to self: this probably will require specific `platform-config` crates to exist, that can be shared between the actual platform crate and the platform crate's build-rs.

rfcs/0199-basic-config.md
```diff
@@ -0,0 +1,91 @@
...
+
+Similarly as new features are added that COULD be supported on all platforms, we will need to add them manually to the configuration items of each crate.
+
+For example: if we have SPI drivers in many platforms, and write a bunch of drivers for external components connected over SPI, we will need to add configuration for each driver to all platforms.
+
+This might be mitigated by grouping these into a common "all spi drivers" configuration type that is included "by composition" into all platform crates, but at some point this will cause friction and/or maintenance burden.
+
+We may want to revisit how to support that in the future.
+
+### "demand based" configuration
```
hawkw, line 81: maybe also worth mentioning runtime device detection/plug-n-play as out of scope here?

rfcs/0199-basic-config.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Compile Time vs Runtime Configuration
+
+At the moment, most configuration will be done at compile time, with configuration values being "baked into" the binary.
+
+However, we should ensure that all configuration values also implement `serde` traits, so that platforms that CAN load configuration at runtime (from an SD Card or other external sources) are able to.
+
+Platforms that can load configuration at runtime may need to provide a "two stage" initialization process: using enough of the "baked in" configuration to initialize the SD Card or other external source, then continuing with the initialization process after merging or replacing the baked in data with the loaded data.
+
+## Inspiration
```
hawkw, line 52: the `rust-osdev/bootloader` crate, a pure-rust bootloader for x86 systems, has a kind of interesting config design where a common config type is defined in the kernel binary, but shared with the bootloader binary and a build tool that links the two. possibly worth looking at for inspiration: https://github.com/rust-osdev/bootloader#usage

rfcs/0199-basic-config.md
```diff
@@ -0,0 +1,91 @@
...
+
+The kernel will provide a single `KernelConfig` type, which is a struct that contains all service and daemon configuration values. We will add functionality to the kernel to take a provided `KernelConfig` type, and automatically launch the configured services.
+
+This capability will allow most platform targets to avoid repeated boilerplate `kernel.initialize(...)` calls at startup.
+
+### Platform crate changes
+
+Platform crates will be expected to provide a `PlatformConfig` type, containing all drivers or daemons that are specific or applicable to this platform (other than the ones defined in the kernel itself). Similar to the kernel, they should also provide a single call that takes a `KernelConfig` and `PlatformConfig`, and automatically launches all kernel and platform services.
+
+One platform may have many targets: For example, the Allwinner D1 platform has three targets currently: the `lichee-rv` dev board, the `mq-pro` dev board, and the `beepy`.
```
hawkw, line 36: it would be nice to see an example of how we plan to abstract over fairly complex stuff like pin configurations in the config system, but that can be added to the RFC in a follow-up.

hawkw (approved):
i'm on board with this design, and interested in seeing where it goes! i left some minor comments

Issue #159, opened 2023-07-15
platform(d1): nice abstraction for a task to get woken by a GPIO interrupt


hawkw:
so we can stop busy-polling the keyboard. it would be nice if the i2c_puppet driver could be still cross-platform and just depend on the cross-platform I2C driver and some kind of GPIO IRQ interface provided by the actual board impls, but idk what that would look like...

hawkw:
@jamesmunns curious if you have any thoughts on this. i might try to take a crack at it soonish

jamesmunns:
@hawkw no specific feedback, other than it would probably be good to follow Embassy's lead on APIs and such. For example, https://docs.embassy.dev/embassy-rp/git/rp2040/gpio/struct.Input.html

We don't need to impl all the methods, but in general a "wait for X edge" is good.

PR #201, opened 2023-07-30
feat(mq-pro): wake `i2c_puppet` driver on IRQs

hawkw:
Depends on #200

This branch changes the MQ Pro board binary to configure an interrupt on the `i2c_puppet` IRQ pin, and wake the driver when that interrupt fires.

I did *not* implement a general-purpose abstraction for GPIO interrupts in this branch (see #159). Therefore, this is all hard-coded and installing this interrupt handler will prevent other code from registering interrupts for any other PB7 pin. However, I figured that "perfect is the enemy of working", and, rather than designing and implementing a more generic GPIO interrupt abstraction in this branch, I just wanted to get something that worked for the keyboard ISR specifically. We can build a nice abstraction for interrupts on GPIO in the future.


platforms/allwinner-d1/boards/src/bin/mq-pro.rs
```diff
@@ -123,3 +128,63 @@ fn main() -> ! {
 
     d1.run()
 }
+
+fn init_i2c_puppet_irq(gpio: &mut d1_pac::GPIO, plic: &mut Plic) -> &'static WaitCell {
+    use mnemos_d1_core::plic::Priority;
+    use d1_pac::Interrupt;
+
+    static I2C_PUPPET_IRQ: WaitCell = WaitCell::new();
```
jamesmunns, line 136: Do you want to have some kind of singleton check here, to ensure this doesn't get called more than once?

platforms/allwinner-d1/boards/src/bin/mq-pro.rs
```diff
@@ -123,3 +128,63 @@ fn main() -> ! {
...
+
+    // configure pin mappings:
+
+    // the i2c_puppet PI_INT line is on the GCLK0/GPIO4 pin on the Pi header,
+    // according to this schematic:
+    // https://github.com/sqfmi/beepy-hardware/blob/d051e65fd95fdadd83154378950b171a001125a8/KiCad/beepberry-schematic-v1.pdf
+    //
+    // according to the MangoPi MQ Pro schematic, that pin is routed to PB7 on
+    // the D1: https://mangopi.org/_media/mq-pro-sch-v12.pdf
+    gpio.pb_cfg0.modify(|_r, w| {
```
jamesmunns, line 146: You might want to enable internal pullups here, depending on whether the rp2040 is in open-drain or push-pull mode.

jamesmunns (approved):
Overall LGTM! Left some nits.

platforms/allwinner-d1/boards/src/bin/mq-pro.rs
hawkw: no, but we could add that, i suppose. i'm not sure if it matters, because i don't think it's a huge issue to set the same bits again...

platforms/allwinner-d1/boards/src/bin/mq-pro.rs
jamesmunns: Ah, just kidding! There are hardware pullups on the beepy board.

Screenshot 2023-07-30 at 19 46 32

hawkw: yup, i was also looking at the schematic and was about to mention that. i'll add a comment.

PR #202, opened 2023-07-30
WIP: Implement basic config

jamesmunns:
Work towards #199


source/config/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+#![cfg_attr(not(any(feature = "use-std", test)), no_std)]
+
+use core::marker::PhantomData;
+
+use serde::{Deserialize, Serialize};
+
+pub struct MnemosConfig<'a, K, P>
+where
+    K: Serialize + Deserialize<'a>,
+    P: Serialize + Deserialize<'a>,
```
hawkw, line 10: hmm, it seems like, if we were to get rid of the bounds on the struct itself, we wouldn't need the `PhantomData` for the deserializer lifetime? Instead, we could just have the bounds on `impl` blocks for this struct, and then we wouldn't need a `PhantomData`.

source/kernel/src/lib.rs
```diff
@@ -294,43 +295,70 @@ impl Kernel {
...
+        } else {
+            let deps = [
+                #[cfg(feature = "serial-trace")]
+                settings.sermux_trace.is_some(),
+                settings.sermux_loopback.is_some(),
+                settings.sermux_hello.is_some(),
+            ];
+
+            if deps.into_iter().any(identity) {
+                tracing::error!("Sermux services configured without sermux! Skipping.");
```
hawkw, line 360: if sermux is disabled, this trace...probably won't go anywhere, unless an alternate tracing sink is enabled...

source/kernel/src/serial_trace.rs
```diff
@@ -378,10 +359,94 @@ impl Subscriber for SerialSubscriber {
...
+            where
+                E: serde::de::Error,
+            {
+                Ok(u8_to_level(v))
+            }
+        }
+
+        d.deserialize_u8(LFVisitor)
+    }
+}
```
hawkw, line 444: hmm...i think representing levels as integers in the SerMux wire proto makes sense, but i don't know if I like representing them as integers in the config file. my preference would be to use [the `FromStr` impl for `LevelFilter`](https://docs.rs/tracing/latest/tracing/level_filters/struct.LevelFilter.html#impl-FromStr-for-LevelFilter), so that level filters in a config file are specified as, e.g.
```toml
[serial_trace]
initial_level = "debug"
```
rather than
```toml
[serial_trace]
initial_level = 1
```

hawkw (commented):
some minor suggestions, LMKWYT

source/kernel/src/lib.rs
jamesmunns: Yeah, it was "trace or panic", and I went with trace. It was a coin toss tho, so if you prefer panic we can do that.

source/kernel/src/serial_trace.rs
jamesmunns: Ahhh, you're probably right. I had my "postcard" brain on, but strings are probably better because this will round trip through json/toml too.

platforms/melpomene/melpo.toml
```diff
@@ -0,0 +1,44 @@
...
+[kernel_svc_cfg]
+keyboard_mux = { max_keyboards = 8, buffer_capacity = 32, sermux_port = 2 }
+serial_mux = { max_ports = 16, max_frame = 512 }
+spawnulator = { capacity = 16 }
+sermux_loopback = { port = 0, buffer_size = 128 }
+
+[kernel_svc_cfg.sermux_hello]
+port = 1
+buffer_size = 32
+message = [ # this sucks
```
jamesmunns, line 14: oh yeah we might want to do something about this, maybe switch to `heapless::String<32>` instead of `[u8; 32]`. Open to ideas here.

platforms/melpomene/melpo-config/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// If this field is None, then the display service will not
+    /// be spawned.
+    pub display: Option<DisplayConfig>,
+
+    /// Forth GUI shell settings
+    ///
+    /// If this field is None, then the shell service will not
+    /// be spawned.
+    pub forth_shell: Option<ForthShell>,
+
```
hawkw, line 30: nit, take it or leave it: this seems not necessarily melpo specific?

platforms/melpomene/melpo-config/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// The maximum kchannel depth for processing messages
+    pub kchannel_depth: usize,
+    /// Incoming TCP buffer size in bytes
+    pub incoming_size: usize,
+    /// Outgoing TCP buffer size in bytes
+    pub outgoing_size: usize,
+    /// Socket addr opened as a simulated serial port
+    ///
+    /// For example: "127.0.0.1:9999"
+    pub socket_addr: String,
```
hawkw, line 47: nit: consider just making this a `SocketAddr`, which [implements `Deserialize`](https://docs.rs/serde/latest/serde/trait.Deserialize.html#impl-Deserialize%3C'de%3E-for-SocketAddr)?

platforms/melpomene/melpo.toml
```diff
@@ -0,0 +1,39 @@
+[kernel_cfg]
```
hawkw, line 1: nit: personally, i don't know if the `_cfg` suffix is particularly useful for these --- the whole _thing_ is a config file, so i might just call it `[kernel]`...

source/config/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "config"
+version = "0.1.0"
+edition = "2021"
+description = """
+Compile time configuration library. Intended to be used by platform crates
+and their build scripts
+"""
```
hawkw, line 8:
```suggestion
"""
readme = "./README.md"
license = "MIT OR Apache-2.0"
repository = "https://github.com/tosc-rs/mnemos"
homepage = "https://mnemos.dev"
```

source/config/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/// Load the configuration created by `render_project` in a build.rs.
+///
+/// Should be called with the type of your platform specific type
+#[macro_export]
+macro_rules! load_configuration {
+    ($platform: ty) => {{
+        const MELPO_CFG: &[u8] = include_bytes!(env!("MNEMOS_CONFIG"));
+        ::config::runtime::from_postcard::<$platform>(MELPO_CFG)
```
hawkw, line 134: nit: should this be:
```suggestion
        const MNEMOS_CFG: &[u8] = include_bytes!(env!("MNEMOS_CONFIG"));
        ::config::runtime::from_postcard::<$platform>(MNEMOS_CFG)
```
since it's not necessarily Melpo?

hawkw, line 134: this should probably be 
```suggestion
        const MELPO_CFG: &[u8] = include_bytes!(env!("MNEMOS_CONFIG"));
        $crate::runtime::from_postcard::<$platform>(MELPO_CFG)
```
in case the crate dependency is renamed?

source/kernel/src/serial_trace.rs
```diff
@@ -382,6 +363,107 @@ impl Subscriber for SerialSubscriber {
...
+
+    /// Capacity for the trace ring buffer.
+    ///
+    /// Note that *two* buffers of this size will be allocated. One buffer is
+    /// used for the normal trace ring buffer, and another is used for the
+    /// interrupt service routine trace ring buffer.
+    pub tracebuf_capacity: usize,
+
+    /// Initial level filter used if the debug host does not select a max level.
+    #[serde(with = "level_filter")]
```
hawkw, line 387: nit: doesn't  serde have a way to just say "use the `FromStr` impl"?

hawkw (approved):
overall, this looks good!

one question is that it seems like there isn't currently a way to declare defaults when the config is declared, so the config file *has* to specify everything. is this desirable, or should we allow leaving a field blank to let it be defaulted?

platforms/melpomene/melpo-config/src/lib.rs
jamesmunns: This is probably a better idea!

platforms/melpomene/melpo-config/src/lib.rs
hawkw: ah, ew. hmm...

source/kernel/src/forth/mod.rs
```diff
@@ -35,6 +35,48 @@ pub struct Params {
...
+                .stdout_capacity
+                .unwrap_or(Params::DEFAULT_STDOUT_CAPACITY),
+            bag_of_holding_capacity: self
+                .bag_of_holding_capacity
+                .unwrap_or(Params::DEFAULT_BAG_OF_HOLDING_CAPACITY),
+            spawnulator_timeout: self
+                .spawnulator_timeout
+                .unwrap_or(Params::DEFAULT_SPAWNULATOR_TIMEOUT),
+        }
+    }
```
hawkw, line 77: i think there might be an easier way of handling the default values: https://serde.rs/attr-default.html

source/kernel/src/forth/mod.rs
```diff
@@ -212,16 +254,25 @@ impl<'forth> AsyncBuiltins<'forth, MnemosContext> for Dispatcher {
...
 
 impl Params {
+    const DEFAULT_STACK_SIZE: usize = 256;
+    const DEFAULT_INPUT_BUF_SIZE: usize = 256;
+    const DEFAULT_OUTPUT_BUF_SIZE: usize = 256;
+    const DEFAULT_DICTIONARY_SIZE: usize = 4096;
+    const DEFAULT_STDIN_CAPACITY: usize = 1024;
+    const DEFAULT_STDOUT_CAPACITY: usize = 1024;
+    const DEFAULT_BAG_OF_HOLDING_CAPACITY: usize = 16;
+    const DEFAULT_SPAWNULATOR_TIMEOUT: Duration = Duration::from_secs(5);
```
hawkw, line 264: personally, i like making the default values `pub`, so that downstream code can say stuff like 

```rust
Params {
    stack_size: Params::DEFAULT_STACK_SIZE * 2, // double the default stack size
   ...Default::default()
}
```
or whatever. not a big deal though.

PR #204, opened 2023-07-30
feat(trace): tune trace-proto grant/buf sizes

hawkw:
Through extremely rigorous statistical analysis of average traceproto message sizes[^1], I've determined that the sizes of `bbq` send grants and buffers used by the trace-proto subscriber are Way Too Big.

This hasn't caused an issue for us yet because the D1 has a massive, overwhelming amount of RAM that we can basically never use all of (1 GB), but it hurts a lot more on the ESP32-C3, which only has 400 KB of RAM, and a 32 KB heap. The current trace-proto subscriber tracebuf size will result in requesting a heap allocation that's *4 times* the size of the entire heap on the ESP32-C3.

Therefore, this commit makes the trace-proto subscriber's send grants and trace buffers substantially smaller, to make it more ESP32-C3-friendly. I did some testing on the D1 by setting everything to `TRACE` and doing host-side filtering in `crowtty` so I don't actually see the individual trace messages, and I didn't see any "discarded trace" events after running for a while. The new sizes should be okay, but we can embiggen them again if we start seeing dropped traces.

[^1]: Running crowtty in verbose mode for a while and looking at "about how big" the the SerMux packets were

### DIFF
--- a/source/kernel/src/serial_trace.rs
+++ b/source/kernel/src/serial_trace.rs
@@ -137,7 +137,7 @@ impl SerialSubscriber {
         use maitake::time;
         use postcard::accumulator::{CobsAccumulator, FeedResult};
 
-        // we probably won't use 256 whole bytes of cobs yet since all the host
+        // we probably won't use 16 whole bytes of cobs yet since all the host
         // -> target messages are quite small
         let mut cobs_buf: CobsAccumulator<16> = CobsAccumulator::new();
         let mut read_level = |rgr: bbq::GrantR| {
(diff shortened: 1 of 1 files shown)


jamesmunns:
Just a note that hopefully we can For Real fix this once #202 lands.

jamesmunns (approved):

hawkw:
> Just a note that hopefully we can For Real fix this once #202 lands.

well, sort of. the default sizes will be configurable (and I was already making them much smaller on the c3), but the sizes of the _write grants_ that we were requesting into those `bbq`s were hardcoded, and they were _way_ too big. asking for a 1KB write grant for every event limits how small we can actually make the `bbq`, and that needed to be fixed as well.

PR #209, opened 2023-07-31
chore(esp32c3): add `core` crate in main workspace 

hawkw:
This change moves most of the driver and initialization code currently
in `platforms/esp32c3-buddy` into a new `platforms/esp32c3/core` crate,
which is part of the main workspace, with only the actual bins built for
the target binaries in their own workspace.

This allows most of the driver code to participate in stuff like `cargo
check` and `cargo fmt` runs in the main workspace, which should make our
lives a bit easier. It was necessary to ensure that incompatible
features of the `critical-section` crate were not enabled in the main
workspace, so I did some tweaking of the Cargo features to ensure that
we weren't enabling an incompatible configuration in core crates.

I also fixed some Clippy lints that immediately showed up in VS Code once
the C3 code was moved into the main workspace.


jamesmunns (approved):

hawkw:
Hmm, it looks like we can't get rid of the dependency on `esp-riscv-rt` in `esp32c3-hal`, so any code that uses the HAL will always include RISC-V assembly in its build. This means that having code that uses the HAL in the main workspace will cause build failures while trying to build for the host's target (assuming the build host is not RISC-V): https://github.com/tosc-rs/mnemos/actions/runs/5717974422/job/15492976270?pr=209

So, I guess this just straight up _won't work_. Oh well.

hawkw:
We could probably do something like this if the `esp32c3-core` crate depended on the [`esp32c3` PAC crate](https://crates.io/crates/esp32c3) instead of `esp32c3-hal`, which would let us just use the PAC types without compiling the inline assembly in the `rt` crate, which would only be compiled when building for the target. I don't know if changing the dependencies like this would really be worth the effort, though.

PR #211, opened 2023-07-31
docs: populate Cargo metadata for every crate

hawkw:
Oranda will use this to generate GitHub links on the various crate pages.

### DIFF
--- a/oranda.json
+++ b/oranda.json
@@ -2,6 +2,9 @@
     "project": {
         "name": "Mnemos Project Overview",
         "description": "This page provides an overview of the entire MnemOS project. The other boxes on this page provide an overview of the individual crates that make up MnemOS.",
-        "readme_path": "./README.md"
+        "readme_path": "./README.md",
+        "repository": "https://github.com/tosc-rs/mnemos",
+        "homepage": "https://mnemos.dev",
+        "license": "MIT OR Apache-2.0"
     }
 }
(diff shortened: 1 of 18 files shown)


jamesmunns:
So, two points:

* I actually *unpopulated* a lot of the repo links, partially because they were wrong, and partially because it seemed a little silly for oranda to link to the git repo for every project (where all links go to the git repo)
* In general, you are welcome to add yourself to any developers list, and/or we can switch to a "the mnemos project developers" or similar if you and/or others prefer.

If the change looks fine to you locally, I'm not strongly against it, and happy to merge.

jamesmunns (approved):

hawkw:
> * I actually _unpopulated_ a lot of the repo links, partially because they were wrong, and partially because it seemed a little silly for oranda to link to the git repo for every project (where all links go to the git repo)

hmm, personally, I think it's nice to let Oranda generate the "check out our GitHub repo" banner at the top of every page. otherwise, it's a bit hard to even _find_ the GitHub if you're just looking at the docs...but, I agree that having a "github" link on the index page for every project in the index is a bit silly. it might be worth seeing if we can make an upstream change so that Oranda will just generate the banner on its index page, if every project has the same repo URL?

incidentally, we also need to at least have "repository" metadata for the root project in order to enable their nice changelog page generation. so we could remove it from all the sub-crates, if you prefer, but i'd like the changelog page....

jamesmunns:
Let's roll with this change for now :)

PR #216, opened 2023-08-02
platform(x86_64): initial x86_64 bringup

hawkw:
Depends on #219.

This branch adds an initial implementation of an x86_64 platform target for
mnemOS, using the [`mycelium` x86 HAL][hal]. Currently, I haven't implemented
drivers for the UART, keyboard, or emb-display service (using the framebuffer as
the display), so we don't have SerMux, Forth, trace-proto, or anything else
interesting. However, what we *do* have is a basic bootable image with a timer,
allocator, and rudimentary kernel run loop, and --- as you can see here --- it
works:

image

I've also added new `just` commands for building x86_64 images and running them
in QEMU for development.

[hal]: https://github.com/hawkw/mycelium/tree/main/hal-x86_64


platforms/x86-64/core/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+
+pub mod acpi;
+pub mod interrupt;
+
+pub fn init(bootinfo: &impl BootInfo, rsdp_addr: Option<PAddr>) -> &'static Kernel {
+    // TODO: init early tracing?
+    interrupt::enable_exceptions();
+    bootinfo.init_paging();
+
+    // TODO: init allocator!
```
hawkw, line 21: the bootinfo will give us a memory map we can use to populate a kernel allocator. this should provide something for `mnemos_alloc` but will also need to implement the [`page::Alloc` trait from mycelium hal](https://mycelium.elizas.website/hal_x86_64/mm/page/trait.alloc).

see https://github.com/hawkw/mycelium/blob/1f125194902cd4970b72eab0aa1d85d1b6ec1489/src/lib.rs#L118-L152

platforms/x86-64/core/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+    cpu::{local::LocalKey, wait_for_interrupt},
+    mm,
+};
+use kernel::{mnemos_alloc::containers::Box, Kernel, KernelSettings};
+
+pub mod acpi;
+pub mod interrupt;
+
+pub fn init(bootinfo: &impl BootInfo, rsdp_addr: Option<PAddr>) -> &'static Kernel {
+    // TODO: init early tracing?
```
hawkw, line 17: would be nice to have framebuffer and/or UART tracing before the runtime comes up...

platforms/x86-64/core/src/lib.rs
hawkw: if we want to init tracing before we init the allocator, we will probably need a small bump region for the `tracing` subscriber's single arc allocation.

platforms/x86-64/core/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+    interrupt::enable_exceptions();
+    bootinfo.init_paging();
+
+    // TODO: init allocator!
+
+    // TODO: PCI?
```
hawkw, line 23: probably punt on PCI for the initial bringup branch...

platforms/x86-64/core/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+        };
+
+        unsafe {
+            Box::into_raw(Kernel::new(settings).expect("cannot initialize kernel"))
+                .as_ref()
+                .unwrap()
+        }
+    };
+
+    // TODO: spawn drivers (UART, keyboard, ...)
```
hawkw, line 46: most of this isn't needed for the initial bringup branch but we probably need either `emb_display` (writing directly to framebuf) or UART in order to prove we are alive. mycelium's `hal-x86_64` crate has modules for the [framebuffer](https://mycelium.elizas.website/hal_x86_64/framebuffer/) as well as [the 16550 UART](https://mycelium.elizas.website/hal_x86_64/serial/), which will need to be wrapped with mnemOS-style driver services eventually (they would implement the [`EmbDisplayService`](https://mnemos-dev.jamesmunns.com/doc/kernel/services/emb_display/) and [`SimpleSerialService`](https://mnemos-dev.jamesmunns.com/doc/kernel/services/simple_serial/), respectively).

platforms/x86-64/core/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+    k
+}
+
+pub fn run(bootinfo: &impl BootInfo, k: &'static Kernel) -> ! {
```
hawkw, line 50: this will probably also need to take some kind of "CPU core context" eventually when we actually get to SMP support...

hawkw (commented):
@saleemrashid, if you want to work on this, feel free to pick up from where i've left off! i left some notes to help you get started.

we'll want to add some bin crates in `platform/x86-64` for actually building mnemOS binaries with various bootloaders. i think the `rust-osdev/bootloader` crate is nice for testing in e.g. QEMU and i already have [code from Mycelium we can copy](https://github.com/hawkw/mycelium/blob/main/src/arch/x86_64/boot.rs) for implementing `hal-x86-64`'s bootinfo trait for `rust-osdev/bootloader`. we can steal the build process from [Mycelium's stupid, overengineered build tool](https://github.com/hawkw/mycelium/blob/main/src/arch/x86_64/boot.rs), or by following [the `bootloader` crate's documentation](https://github.com/rust-osdev/bootloader/blob/main/docs/create-disk-image.md). any other bootloaders we want to support will need to either [chainload into bootloader](https://github.com/rust-osdev/bootloader/blob/main/docs/chainloading.md) to populate its bootinfo, or have their own way of providing the [`hal-core::BootInfo` trait](https://mycelium.elizas.website/hal_core/boot/trait.bootinfo), implemented in that bootloader's bin target.

PR #217, opened 2023-08-03
WIP: Start implementing config for the d1

jamesmunns:


platforms/allwinner-d1/build.rs
```diff
@@ -0,0 +1,36 @@
...
+    println!("cargo:rerun-if-changed=memory.x");
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rustc-link-arg=-Tmemory.x");
+    println!("cargo:rustc-link-arg=-Tlink.x");
+
+    let lichee_rv = cfg!(feature = "lichee-rv");
+    let mq_pro = cfg!(feature = "mq-pro");
+    let beepy = cfg!(feature = "beepy");
+
+    let name = match (lichee_rv, mq_pro, beepy) {
```
hawkw, line 27: hmm, what does the "beepy" cfg indicate here? i kind of figured that the "beepy" feature flag would control whether or not the board is attached to a Beepy (e.g. should spawn the i2c_puppet and beepy display drivers), rather than being a separate target entirely?

platforms/allwinner-d1/d1-config/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl I2cConfiguration {
+    const fn default_mapping() -> Mapping {
+        Mapping::LicheeRvTwi0
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub enum Mapping {
+    LicheeRvTwi0,
```
hawkw, line 30: this is backwards; it's TWI2 on the lichee rv and TWI0 on the mq pro.

jamesmunns: Good catch! Will fix.

hawkw:
this is dope as hell btw, nice work! :D

platforms/allwinner-d1/build.rs
jamesmunns: So I was using "one cfg per toml", it's still probably tbd to determine how we set which config to use. It will probably be driven by `just`, but it could be a feature flag or an env var or something.

Right now the options translate to:

* "lichee-rv in a dock"
* "bare mq-pro board"
* "mq-pro in a beepy"

platforms/allwinner-d1/d1-config/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+    MangoPiTwi2,
+}
```
hawkw, line 32: all the D1s have TWI peripherals 0-3, so maybe this should be 
```suggestion
#[derive(Debug, Serialize, Deserialize)]
pub enum Twi {
    Twi0,
    // Twi1,
    Twi2,
    // Twi3,
}
```
and add a separate thing to configure which pins that twi is mapped to? but maybe that's just unnecessary complexity at this point...

platforms/allwinner-d1/build.rs
hawkw: okay, yeah, that makes sense for now. this will have to change when we do a beepy shield for the RV, though. then it will be a matrix of `(board, is_beepy)`

hawkw: okay, yeah, that makes sense for now. this will have to change when we do a beepy shield for the RV, though. then it will be a matrix of `(board, is_beepy)`

jamesmunns: @hawkw I sort of disagree, we don't *really* have a way to do "overlays", so for now, every "configuration" gets its own toml file. So then we would have:

* lichee-rv-dock.toml
* lichee-rv-beepy.toml
* mq-pro-bare.toml
* mq-pro-beepy.toml

platforms/allwinner-d1/d1-config/src/lib.rs
jamesmunns: Yeah, since we don't actually support full mapping, it would be something like (pin numbers made up):

* Twi0_PB1_PB7
* Twi2_PG13_PG15

Since they can't really be decoupled.

jamesmunns:
(deleted some comments from @hawkw because https://github.com/tosc-rs/mnemos/pull/217#discussion_r1282647920 got triple-posted somehow.

PR #218, opened 2023-08-03
chore: add Nix dev shell

hawkw:
I've been using this configuration for a while, so I thought I'd commit it to make life easier for any other folks using Nix. I started working on rewriting the configuration as a Nix flake but ran into some issues, so I might continue messing with that later, but this works for now.


jamesmunns (approved):
I don't know anything about nix, but if it makes your life easier, go off :D

PR #219, opened 2023-08-09
chore:  move to One Big Cargo Workspace

hawkw:
This commit changes the Cargo workspace setup to put all crates in One
Big Workspace, rather than having separate workspaces for some targets.
We now use the `per-package-target` unstable cargo feature to build
different crates for different targets. This means that `cargo` commands
in the root workspace now work without requiring the user to `cd` into a
particular directory to build a platform target --- for example, I can
now run:

```shell
# in the repo root directory
$ cargo build -p mnemos-d1 --bin mq-pro
```

and build a MnemOS binary for the MQ Pro, without having to `cd` into
the MQ Pro directory.

This is also necessary in order to make the `x86_64` build process added
in PR #216 work, since it relies on cargo artifact dependencies, which
appear not to work across workspaces.

One issue is that `cargo build --workspace` (and `check --workspace`,
etc) still does not work correctly, due to some [weird][1] [issues][2]
with feature unification which I don't entirely understand. However, as
a workaround, I've changed the workspace Cargo.toml to add a
[`default-members` field][3], so that running `cargo build` or `cargo check` 
_without_ `--workspace` will build the subset of crates in the
`default-members` key. This way, `cargo {build, check, etc}` in the repo
root will do something reasonable by default, and the actual platform
targets can be built/checked with `cargo $WHATEVER --package $CRATE`.
IMO, this is still substantially nicer than having a bunch of separate
workspaces.

[1]: https://github.com/ia0/data-encoding/issues/47
[2]: https://github.com/bincode-org/bincode/issues/556
[3]: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-default-members-field


jamesmunns (approved):
Overall LGTM, it would be nice to document *somewhere* the nightly cargo trickery we are using and why, and link to the related tracking issues so we know when we can get rid of them, but if it still works and produces valid binaries for the various targets, no objections here.

platforms/esp32c3-buddy/.cargo/config.toml
```diff
@@ -1,21 +0,0 @@
-[target.riscv32imac-unknown-none-elf]
-runner = "espflash flash --monitor"
-
-[build]
-rustflags = [
-  "-C", "link-arg=-Tlinkall.x",
```
jamesmunns, line 6: @hawkw where did this end up living? I can't see it somewhere else, and the linker scripts need to be there.

jamesmunns: Nevermind found it

hawkw:
> Overall LGTM, it would be nice to document _somewhere_ the nightly cargo trickery we are using and why, and link to the related tracking issues so we know when we can get rid of them, but if it still works and produces valid binaries for the various targets, no objections here.

Mostly, we are just using features that haven't yet been stabilized. These aren't workarounds for *bugs*, per se. I'm happy to add links to the tracking issues for those features.

hawkw:
@jamesmunns https://github.com/tosc-rs/mnemos/pull/219/commits/069c07aca23559e3d0f1d8a0e3e217d33e5b17a1 adds comments explaining the various unstable features & stuff.

PR #230, opened 2023-08-15
bug(docs): Add netlify dry run in CI

jamesmunns:
This PR changes the netlify CI to use `just doc`, and adds a GH action to run these steps
to give us an advance warning if the netlify job is likely to fail.


jamesmunns:
Merging to check whether it fixes netlify

PR #239, opened 2023-08-18
feat(kernel): add a generic GPIO pin abstraction

hawkw:


hawkw:
@jamesmunns WDYT of this?

hawkw:
belatedly, i realized that https://docs.rs/embedded-hal-async/0.2.0-alpha.2/embedded_hal_async/digital/trait.Wait.html also exists for waiting on pin IRQs, so...we might be better off just using that? i think my API design is slightly nicer but this is compatible with other` embedded-hal-async` crates

PR #240, opened 2023-08-23
docs: add script to add RFCs to mdBook

hawkw:
This branch adds a Python script that appends the RFCs in the `rfcs` dir to the mdBook.

I'm not 100% convinced this is the "best approach", since the script has to be run before building the mdBook. Alternatively, we could just make the RFCs dir a subdirectory of the mdBook, and require PRs that add new RFCs to add them to the index as well. That requires more effort when adding an RFC, and means they can no longer live in a top-level `rfcs/` directory. But, on the other hand, it means you don't have to run a big wad of random janky python before you can generate the book.

@jamesmunns WDYT?


jamesmunns:
No objections, would it be worth it to add it to the "run CI" script in the `scripts/` folder so it gets run automatically on docs builds?

hawkw:
> No objections, would it be worth it to add it to the "run CI" script in the `scripts/` folder so it gets run automatically on docs builds?

yeah, i haven't done the plumbing to add it to CI, or added docs; wanted to make sure we felt good about this approach first. happy to add it if we feel good about this way of generating the RFCs!

jamesmunns (approved):

PR #242, opened 2023-08-25
chore(nix): replace nix-shell with flake, add Oranda

hawkw:
Flakes are the Shiny New Way of doing stuff in Nix-land, so this branch
replaces the current `default.nix`/`shell.nix` with a flake. Using a
flake also allows us to depend on an up-to-date version of Oranda, which
is provided as a flake --- the version in nixpkgs is v0.2.0, rather than
v0.3.1, so depending on Oranda's flake gets us the proper version in the
dev shell.

Eventually, we can make this much fancier, with e.g. different flake
`app` derivations for the various host tools, so that they can be used
with `nix run`. I might mess with that later. But for now, this gets us
a dev shell that's equally nice to the previous one and also has Oranda
in it...


hawkw:
@jamesmunns i'm just gonna merge this as i assume you don't care about it

jamesmunns:
@hawkw you're the box of nix round these parts!

jamesmunns (approved):

PR #258, opened 2023-08-28
feat(kernel): add listeners to services

hawkw:
Currently, when a task (or userspace) wishes to connect to a `RegisteredDriver`,
it performs a registry lookup, which returns a handle for sending requests to
that service. These handles are essentially producers for a single `KChannel`
which is associated with the service when it's registered, and the service can
use this channel's `KConsumer` to receive a stream of incoming requests.

This approach works fine, but it limits what a service can do with its incoming
request stream. Because all requests arrive on the same `KConsumer`, the server
may not spawn a separate task to handle requests from each client. Similarly, it cannot
route requests to different worker tasks based on which resource the request is
associated with.

This branch changes the way clients communicate with servers by introducing a
_handshake_ to this process. Rather than registering a service by giving the
registry a `KProducer` for a single request channel, the registry now owns a
producer for a channel of _incoming connections_. These connections contain a
`Hello` message sent by the client, of a type defined by the `RegisteredDriver`
implementation for the service. This may be used by the service to determine
what resource the client wants to access, or otherwise route connections to
different tasks. This handshake is completed by the service responding to the
incoming connection with a `KProducer` for a request channel, which is then
returned to the client.

The returned `KProducer` _may_ still point to a single global request channel
for that service, allowing us to continue to implement the existing behavior.
However, the service may alternatively spawn a new task for that connection and
create a new channel for that worker. Or, it may return a different `KProducer`
for a different request channel depending on the `Hello` message, allowing the
server to have separate worker tasks for different resources managed by the same
service, and route the connection to one of those workers depending on which
resource was requested in the `Hello` message.

An adapter, `into_request_stream`, is provided, to make it easier for services
which want the current behavior of one global stream of requests to have that.
Currently, all the existing services use this, but in the future, many of our
existing services can be rewritten to use the new handshake mechanism more
effectively. I'd also like to do some additional refactoring, but I wanted to
get a working PR up first.

### DIFF
--- a/platforms/beepy/src/i2c_puppet.rs
+++ b/platforms/beepy/src/i2c_puppet.rs
@@ -132,7 +132,11 @@ impl I2cPuppetClient {
     pub async fn from_registry_no_retry(
         kernel: &'static Kernel,
     ) -> Result<Self, registry::ConnectError<I2cPuppetService>> {
-        let handle = kernel.registry().await.get::<I2cPuppetService>().await?;
+        let handle = kernel
+            .registry()
+            .await
+            .connect::<I2cPuppetService>()
+            .await?;
 
         Ok(I2cPuppetClient {
             handle,
--- a/platforms/beepy/src/i2c_puppet.rs
+++ b/platforms/beepy/src/i2c_puppet.rs
@@ -111,8 +111,11 @@ impl I2cPuppetClient {
     pub async fn from_registry(kernel: &'static Kernel) -> Self {
         loop {
             match I2cPuppetClient::from_registry_no_retry(kernel).await {
-                Some(port) => return port,
-                None => {
+                Ok(port) => return port,
+                Err(registry::ConnectError::Rejected(_)) => {
+                    unreachable!("the KeyboardMuxService does not return connect errors!")
+                }
+                Err(_) => {
                     // I2C probably isn't registered yet. Try again in a bit
                     kernel.sleep(Duration::from_millis(10)).await;
                 }
@@ -126,12 +129,12 @@ impl I2cPuppetClient {
     ///
     /// Prefer [`I2cPuppetClient::from_registry`] unless you will not be spawning one
     /// around the same time as obtaining a client.
-    pub async fn from_registry_no_retry(kernel: &'static Kernel) -> Option<Self> {
-        let handle = kernel
-            .with_registry(|reg| reg.get::<I2cPuppetService>())
-            .await?;
+    pub async fn from_registry_no_retry(
+        kernel: &'static Kernel,
+    ) -> Result<Self, registry::ConnectError<I2cPuppetService>> {
+        let handle = kernel.registry().await.get::<I2cPuppetService>().await?;
 
-        Some(I2cPuppetClient {
+        Ok(I2cPuppetClient {
             handle,
             reply: Reusable::new_async().await,
         })
@@ -227,7 +230,7 @@ impl I2cPuppetClient {
 /// Server implementation for the [`I2cPuppetService`].
 pub struct I2cPuppetServer {
     settings: I2cPuppetSettings,
-    rx: KConsumer<registry::Message<I2cPuppetService>>,
+    rx: registry::listener::RequestStream<I2cPuppetService>,
     i2c: I2cClient,
     subscriptions: FixedVec<KProducer<(KeyStatus, KeyRaw)>>,
     keymux: Option<KeyboardMuxClient>,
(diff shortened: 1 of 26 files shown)


platforms/beepy/src/i2c_puppet.rs
```diff
@@ -109,8 +111,11 @@ impl I2cPuppetClient {
...
             match I2cPuppetClient::from_registry_no_retry(kernel).await {
-                Some(port) => return port,
-                None => {
+                Ok(port) => return port,
+                Err(registry::ConnectError::Rejected(_)) => {
+                    unreachable!("the KeyboardMuxService does not return connect errors!")
+                }
+                Err(_) => {
                     // I2C probably isn't registered yet. Try again in a bit
                     kernel.sleep(Duration::from_millis(10)).await;
```
jamesmunns, line 120: I wonder if we should give the registry a waitqueue for "new service registered".

source/kernel/src/registry/mod.rs
```diff
@@ -62,14 +75,38 @@ pub trait RegisteredDriver {
...
+    /// An initial message sent to the service by a client when establishing a
+    /// connection.
+    ///
+    /// This may be used by the service to route connections to specific
+    /// resources owned by that service, or to determine whether or not the
+    /// connection can be established. If the service does not require initial
+    /// data from the client, this type can be set to [`()`].
+    // XXX(eliza): ideally, we could default `Hello` to () and `ConnectError` to
+    // `Infallible`...do we want to do that? it requires a nightly feature.
+    type Hello: 'static;
```
jamesmunns, line 87: Nit, we might want to give this a name like "Introduction" or "ConnectDetails" or something, but we can bikeshed later.

source/kernel/src/registry/mod.rs
```diff
@@ -62,14 +75,38 @@ pub trait RegisteredDriver {
...
+    ///
+    /// If the service does not reject connections, this should be set to
+    /// [`core::convert::Infallible`].
+    type ConnectError: 'static;
+
     /// This is the UUID of the driver service
     const UUID: Uuid;
 
     /// Get the type_id used to make sure that driver instances are correctly typed.
     /// Corresponds to the same type ID as `(Self::Request, Self::Response, Self::Error)`
```
jamesmunns, line 100: update comment here too please

source/kernel/src/registry/mod.rs
```diff
@@ -229,6 +266,29 @@ pub enum RegistrationError {
...
 }
 
+/// Errors returned by [`Registry::connect`] and
+/// [`Registry::connect_with_hello`].
+pub enum ConnectError<D: RegisteredDriver> {
+    /// No [`RegisteredDriver`] of this type was found!
+    NotFound,
+    /// The remote [`RegisteredDriver`] rejected the connection.
+    Rejected(D::ConnectError),
+    /// The remote [`RegisteredDriver`] has been registered, but the service
```
jamesmunns, line 276: I wonder if we could check this in the registry (awaiting channel closure?), tho we can't remove services anyway yet.

source/kernel/src/registry/mod.rs
```diff
@@ -284,26 +344,44 @@ pub struct KernelHandle<RD: RegisteredDriver> {
...
-type ErasedDeserHandler = unsafe fn(
+type ErasedReqDeser = unsafe fn(
     UserRequest<'_>,
     &ErasedKProducer,
     &bbq::MpscProducer,
     ServiceId,
     ClientId,
 ) -> Result<(), UserHandlerError>;
 
+type ErasedHandshake = unsafe fn(
```
jamesmunns, line 355: Honestly, if we want to just yank all of the userspace stuff for now, we've already removed the k2u/u2k rings, and this will all be redesigned to be interface-like later.

jamesmunns (approved):
Made some notes, happy to merge this and follow up. Thank you!

platforms/beepy/src/i2c_puppet.rs
hawkw: that's a great idea actually, hmm!

source/kernel/src/registry/mod.rs
hawkw: i stole the name from TLS handshakes, which start with a  `ClientHello` message followed by a `ServerHello` message. But, if you don't like it, i'm happy to change it later...

platforms/beepy/src/i2c_puppet.rs
hawkw: would prefer to address that in a follow-up, though.

source/kernel/src/registry/mod.rs
hawkw: my motivation for actually writing the userspace code was mostly that i wanted to validate that the new design didn't completely break our type erasure mechanism, and that it was possible to do something like this for a serialization based service. i'm happy to remove this code and add something similar back in later, but i have a slight preference for keeping it so that we can build on top of it when integrating the new serializing channel. fundamentally up to you, though.

PR #262, opened 2023-08-29
feat(kernel/daemons/shell): Smarter redraw debouncing

### LINKED ISSUE #255: Efficiency: graphical shell prevents sleep
currently `mnemos_kernel::daemons::shells::graphical_shell_mono` polls at 20Hz, preventing the OS to enter a WFI state

jamesmunns:
This PR introduces smarter debouncing of GUI shell redraws, by first waiting "forever",
then waiting with a timeout.

Closes #255

### DIFF
--- a/platforms/melpomene/src/main.rs
+++ b/platforms/melpomene/src/main.rs
@@ -94,8 +94,11 @@ async fn kernel_entry() {
         tracing::warn!("Not spawning TCP UART server!");
     }
 
+    let mut debounce_period = Duration::from_millis(50);
+
     // Spawn the graphics driver
     if config.platform.display.enabled {
+        debounce_period = Duration::from_secs(1) / config.platform.display.frames_per_second as u32;
         k.initialize(async move {
             SimDisplay::register(
                 k,
@@ -120,6 +123,7 @@ async fn kernel_entry() {
         let forth_shell = config.platform.forth_shell;
         guish.capacity = forth_shell.capacity;
         guish.forth_settings = forth_shell.params;
+        guish.redraw_debounce = debounce_period;
         k.initialize(graphical_shell_mono(k, guish)).unwrap();
     } else {
         tracing::warn!("Not spawning forth GUI shell!");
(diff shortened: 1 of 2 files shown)


platforms/melpomene/src/main.rs
```diff
@@ -94,8 +94,11 @@ async fn kernel_entry() {
         tracing::warn!("Not spawning TCP UART server!");
     }
 
+    let mut debounce_period = Duration::from_millis(50);
+
     // Spawn the graphics driver
     if config.platform.display.enabled {
+        debounce_period = Duration::from_secs(1) / config.platform.display.frames_per_second as u32;
```
hawkw, line 101: it would be neat if there was eventually a way to do this generically by having the shell ask the display what its frame rate is...but we should save that for a future PR, probably.

source/kernel/src/daemons/shells.rs
```diff
@@ -222,94 +227,129 @@ pub async fn graphical_shell_mono(k: &'static Kernel, settings: GraphicalShellSe
...
+        io_poll(PollStyle::OneShot, &mut keyboard, &mut rline, &tid_io).await;
 
+        // SOMETHING happened, so now try and grab as many things as possible
+        // until the debounce timer expires
         let _ = k
             .timeout(
-                interval,
-                io_poll(&mut any, &mut keyboard, &mut rline, &tid_io),
+                redraw_debounce,
+                io_poll(PollStyle::Forever, &mut keyboard, &mut rline, &tid_io),
```
hawkw, line 243: nit, take it or leave it: would it be simpler to just make this
```suggestion
                async move {
                    loop {
                        io_poll(&mut keyboard, &mut rline, &tid_io),
                    }
                 }
```
instead of passing the poll style into the `io_poll` function? or was there some reason that didn't work?

hawkw (approved):
looks great!

source/kernel/src/daemons/shells.rs
jamesmunns: None, other than a vague preference to use async functions over blocks.

PR #263, opened 2023-08-30
chore: pomelo workspace checks unfucking

hawkw:
This branch makes two changes:

1. Turns out I had forgotten to add `pomelo` to the Just commands that run a
   check for all crates, like `just check`, `just clippy`, and `just docs`. So,
   it was never actually being checked, on CI or in VS Code (MY BAD LMAO).
   Therefore, this branch adds it to those recipes so it actually gets checked.

2. Once we actually include `pomelo` in `just check`, it turns out that the
   build is fucked, because Cargo doesn't realize what target to build `pomelo`
   for. Thus, I've also added a missing `forced-target="wasm32-unknown-unknown`
   to `pomelo`'s `Cargo.toml`, which should unfuck it.

3. Also I fixed an "unnecessary mut" warning, because now that we're actually
    linting `pomelo` on CI, it turns out that it, uh, has lints. Whoops.

### DIFF
--- a/justfile
+++ b/justfile
@@ -59,7 +59,8 @@ _espbuddy_pkg := "mnemos-esp32c3-buddy"
 _x86_pkg := "mnemos-x86_64-bootloader"
 _x86_bootloader_pkg := "mnemos-x86_64-bootloader"
 
-_pomelo_index_path := "platforms"/"pomelo"/"index.html"
+_pomelo_pkg := "pomelo"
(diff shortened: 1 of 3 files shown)


jamesmunns (approved):

hawkw:
well, now CI fails for an actually legitimate reason, because pomelo has one warning in it: https://github.com/tosc-rs/mnemos/actions/runs/6018848174/job/16327759771?pr=263#step:8:1186

PR #264, opened 2023-08-30
chore(justfile): remove `-Dwarnings` from Clippy

hawkw:
Currently, the local Clippy run will deny warnings even when not on CI. This is a bit annoying, since we configure VS Code to run `just clippy`, and denying warnings means all warnings show up as errors rather than warnings in VS Code's "Problems" view.

### DIFF
--- a/justfile
+++ b/justfile
@@ -32,9 +32,7 @@ _rustflags := env_var_or_default("RUSTFLAGS", "")
 
 # If we're running in Github Actions and cargo-action-fmt is installed, then add
 # a command suffix that formats errors.
-#
-# Clippy version also gets -Dwarnings.
-_fmt_clippy := if env_var_or_default("GITHUB_ACTIONS", "") != "true" { "-- -Dwarnings" } else {
+_fmt_clippy := if env_var_or_default("GITHUB_ACTIONS", "") != "true" { "" } else {
     ```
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -32,9 +32,7 @@ _rustflags := env_var_or_default("RUSTFLAGS", "")
 
 # If we're running in Github Actions and cargo-action-fmt is installed, then add
 # a command suffix that formats errors.
-#
-# Clippy version also gets -Dwarnings.
-_fmt_clippy := if env_var_or_default("GITHUB_ACTIONS", "") != "true" { "-- -Dwarnings" } else {
+_fmt_clippy := if env_var_or_default("GITHUB_ACTIONS", "") != "true" { "" } else {
```
jamesmunns, line 35: Do we have some way of handling this to actually deny clippy warnings in CI?

hawkw: if `GITHUB_ACTIONS=true` it goes into the `else`, which is what sets `-Dwarnings`

hawkw: it's the case where `GITHUB_ACTIONS != "true"` where we no longer deny warnings

jamesmunns: ah, thank you, lacking reading comprehension.

jamesmunns (approved):

PR #267, opened 2023-09-03
feat(kernel): `RwLock` registry and add helpers

hawkw:
If we replace the `u32` counter with an `AtomicU32`, we can make the
various `connect` methods on the registry borrow it immutably. This
allows us to store it in a `RwLock` rather than a `Mutex`, with the
write lock only being needed to register a new driver. I've done this,
as well as replacing `Kernel::with_registry` with `Kernel::registry` and
`Kernel::registry_mut` accessors.

Additionally, I've added new `Registry::bind` and `Registry::bind_konly`
helpers. These create a new listener for a service and register it with
the registry. This can be implemented using existing code, but now it
can be done in a single function call. One thing to note is that doing
this in one call means that the service is added to the registry
_before_ the server task is actually spawned, because we need to create
a listener to give to the server task. However, I don't actually believe
that's a problem, since the the listener is a queue. If some other task
connects to the service before the server is actually running, that's
fine; the request will just go in the queue for use later. This also has
the substantial advantage that server tasks don't get spawned if there's
already a registration for that service. With the current design, we
spawn tasks first and _then_ register, and if the registration fails,
those tasks are still running and nothing will get rid of them. So,
using `bind` also avoids a potential task leak when spawning a duplicate
service.

### DIFF
--- a/platforms/allwinner-d1/core/src/drivers/sharp_display.rs
+++ b/platforms/allwinner-d1/core/src/drivers/sharp_display.rs
@@ -111,7 +111,7 @@ impl SharpDisplay {
         kernel.spawn(draw.draw_run()).await;
 
         kernel
-            .with_registry(|reg| reg.register_konly::<EmbDisplayService>(registration))
+            .register_konly::<EmbDisplayService>(registration)
             .await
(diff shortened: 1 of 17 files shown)


jamesmunns (approved):
hell yeah

PR #268, opened 2023-09-03
feat(kernel): wake tasks on service registration

hawkw:
Depends on #267

Currently, the `Registry::connect_*` methods return an error immediately
if the requested service is not found in the registry. Most client types
implement a `from_registry` method which retry in a loop when the
requested service is not found in the registry. These methods will wait
for a fixed (short) amount of time and then try again until the registry
returns the service.

This approach is quite inefficient, as we have to run a bunch of retry
loops that keep trying to access a service that may not be there. This
may happen several times before the service actually is registered,
especially when registering a service requires connecting to another
service.

This branch improves the efficiency of waiting for a service to be
registered. Now, rather than retrying with a fixed-duration sleep, we
instead have the `Registry` own a `WaitCell` which is woken whenever a
new service is registered. This wakes all takes potentially waiting to
connect, allowing them to re-check whether the service they want is in
the registry. This idea was initially proposed by @jamesmunns in a
[comment] on PR #259

Connections are now established using either `Registry::connect`, which
retries whenever a new service is registered, or `Registry::try_connect`
which never retries.

Additionally, we now have the capacity to indicate that a service is not
found *and* that the registry is full, by closing the `WaitCell`. In
this case, retrying will never succeed, because the registry is full and
if the service isn't already there, it will never be added. In this
case, the retrying methods will also return an error, rather than never
completing, so we avoid a potential task leak.

In order to make this change, we need to move the `RwLock` from being
around the entire `Registry` to being inside the registry, around
`items`. This allows the `WaitCell` to be accessed regardless. It also
allows us to shorten the duration for which the lock is held. This
requires changing all methods on `Registry` to take `&self`.
Therefore, I've removed the wrapper methods on `Kernel` for connecting
and registering, since they can now just be called on `kernel.registry`
without a bunch of extra boilerplate for lock management. I've also
simplified the API surface of the registry a bit by removing the
`connect` methods that don't take a `Hello`, and just using
`Registry::connect(())` in those cases. IMO, those methods weren't
really pulling their weight, and they required us to have a method named
`Registry::try_connect_userspace_with_hello` if we were going to add a
non-retrying `connect` variant. Now, we can just have
`Registry::try_connect_userspace`, `Registry::connect_userspace`,
`Registry::connect`, and `Registry::try_connect`, which feels much less
egregious.

[comment]: https://github.com/tosc-rs/mnemos/pull/258#discussion_r1314282797

### DIFF
--- a/source/kernel/src/registry/mod.rs
+++ b/source/kernel/src/registry/mod.rs
@@ -6,6 +6,7 @@ use core::{
 };
 
 use crate::comms::{kchannel, oneshot::Reusable};
+use maitake::sync::{RwLock, WaitQueue};
 use mnemos_alloc::containers::FixedVec;
 use portable_atomic::{AtomicU32, Ordering};
 use postcard::experimental::max_size::MaxSize;
@@ -63,9 +64,9 @@ pub mod known_uuids {
 ///
 /// Typically used with [`Registry::register`] or [`Registry::register_konly`].
 /// A connection to the service can be established using [`Registry::connect`],
-/// [`Registry::connect_with_hello`], or
-/// [`Registry::connect_userspace_with_hello`] (depending on the service), after
-/// the service has been registered..
+/// [`Registry::try_connect`], [`Registry::connect_userspace`], or
+/// [`Registry::try_connect_userspace] (depending on the service), after
+/// the service has been registered.
 pub trait RegisteredDriver {
     /// This is the type of the request sent TO the driver service
     type Request: 'static;
@@ -119,8 +120,9 @@ pub struct RegistryType {
 
 /// The driver registry used by the kernel.
 pub struct Registry {
-    items: FixedVec<RegistryItem>,
+    items: RwLock<FixedVec<RegistryItem>>,
     counter: AtomicU32,
+    service_added: WaitQueue,
 }
 
 // TODO: This probably goes into the ABI crate, here is fine for now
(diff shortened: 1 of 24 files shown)


source/kernel/src/registry/mod.rs
```diff
@@ -709,38 +816,56 @@ impl Registry {
...
+            service_id,
             client_id: ClientId(client_id),
         })
     }
 
-    fn get<RD: RegisteredDriver>(&self) -> Result<&RegistryItem, ConnectError<RD>> {
-        let Some(item) = self.items.as_slice().iter().find(|i| i.key == RD::UUID) else {
+    async fn insert_item(&self, item: RegistryItem) -> Result<(), RegistrationError> {
+        {
+            let mut items = self.items.write().await;
```
jamesmunns, line 842: Add "load bearing scope" notice here too?

jamesmunns (approved):
Again, dope as hell.

source/kernel/src/registry/mod.rs
hawkw: i don't actually think this one is load-bearing --- it's equivalent anyway, since the woken tasks won't get to run until this function returns, and the lock is dropped at the end of the function. this only makes things slightly more efficient in concurrent use-cases, where the wake call might wake a task on another core that might try to grab the lock. but this can't deadlock in either case.

PR #271, opened 2023-09-14
platform(d1): introduce config for the D1

hawkw:
Closes #217

This branch introduces `mnemos-config` support for the Allwinner D1
platform, building on @jamesmunns' work in #217. I've made a few changes
from James' initial implementation. In particular, I've changed how the
board config is determined. Now, rather than using feature flags to
determine what board is being built, we once again use bin targets.
However, each bin target now has a *very* simple `main` function that
simply selects which config file to include. For example, the
`lichee-rv.rs` bin target now just looks like this:

```rust
#![no_std]
#![no_main]

#[allow(non_snake_case)]
#[riscv_rt::entry]
fn main() -> ! {
    let config = mnemos_config::include_config!(mnemos_d1::PlatformConfig, "lichee-rv").unwrap();
    mnemos_d1::kernel_entry(config);
}

```


hawkw:
bleh i guess CI is wedged

PR #277, opened 2023-09-16
Add a safe "into_data()" method to turn `Word`s into `i32`s

jamesmunns:


source/forth3/src/word.rs
```diff
@@ -55,6 +55,11 @@ impl Word {
         }
     }
 
+    #[inline]
+    pub fn into_data(self) -> i32 {
```
hawkw, line 59: tbh, i would be equally fine with just calling this `data()`. but, up to you, it is technically an into conversion... :woman_shrugging:

hawkw (approved):

source/forth3/src/word.rs
jamesmunns: `Word::data(i32)` is already the (safe) constructor, though that might be equally poorly named.

PR #278, opened 2023-09-16
fix(d1): put back missing `serial-trace` feature

hawkw:
Looks like the feature flag was lost in #271, so the D1 binaries no
longer emit traceproto output. I believe this happened when combining
the `mnemos-d1` crate (which used to enable the feature) with the
`mnemos-d1-core` crate (which didn't enable the feature).

This commit puts it back. :)

I've also modified the Just recipes for building D1 images to allow
passing additional args to Cargo. This way, users of the Just recipes
can now pass `--no-default-features --features ...` to pick which features
are enabled when building. This was necessary for testing what happened
when the board config had a `serial-trace` section but the feature flag was
disabled. But, passing extra cargo args may also be useful in other 
situations.

### DIFF
--- a/justfile
+++ b/justfile
@@ -130,30 +130,33 @@ fmt:
     {{ _cargo }} fmt --package {{ _mn_pkg }}
 
 # build a Mnemos binary for the Allwinner D1
-build-d1 board='mq-pro': (_get-cargo-command "objcopy" "cargo-binutils")
+build-d1 board='mq-pro' *CARGO_ARGS='': (_get-cargo-command "objcopy" "cargo-binutils")
     {{ _cargo }} build \
         --profile {{ profile }} \
         --package {{ _d1_pkg }} \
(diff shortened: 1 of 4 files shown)


justfile
```diff
@@ -130,30 +130,33 @@ fmt:
     {{ _cargo }} fmt --package {{ _mn_pkg }}
 
 # build a Mnemos binary for the Allwinner D1
-build-d1 board='mq-pro': (_get-cargo-command "objcopy" "cargo-binutils")
+build-d1 board='mq-pro' *CARGO_ARGS='': (_get-cargo-command "objcopy" "cargo-binutils")
```
hawkw, line 133: at some point we *may* just want to start turning these just recipes into proper `cargo xtask`s with `mn`, so that we can pass more complex CLI args to them...but i didn't want to do that as part of this change.

PR #279, opened 2023-09-16
feat(crowtty): more readable tracing messages

hawkw:
Depends on #278.

This commit makes a very small change to how the crowtty traceproto
output formats an event's `message` field. The message is now printed
first, and (when the terminal supports ANSI formatting escapes), it's
bolded, so that it stands out a bit from other fields and metadata.
Also, I've changed `crowtty` so that the message field is always printed
first, rather than ordering it alphabetically by field name. This should
improve the consistency of log lines and hopefully make them a bit
easier to read.

For example:

image


jamesmunns (approved):
Always down for PRs that bring joy

PR #280, opened 2023-09-16
feat(d1): allocate DMA channels from the DMAC

hawkw:
Currently, the assignment of D1 DMAC channels to drivers is somewhat ad-hoc. We hard-code channel 0 as belonging to the UART, and channel 1 as belonging to the SPI_DBI driver, with the interrupt handler for DMAC IRQs specifically dispatching those channel's IRQs to those drivers manually. 

There are some issues with this. The exclusive use of DMA channels is not enforced by the current API, because you can (unsafely) `summon_channel` whenever you want and use a DMA channel that someone else is already using. Furthermore, adding new drivers that also use DMA channels requires hard-coding those additional channel IRQs in the interrupt handler, which must then be kept in sync with the channels actually used by the driver. Also, when a driver wants to wait asynchronously on a DMA transfer, it has to ensure that the DMAC ISR will wake its hard-coded `WaitCell`, and the driver code is responsible for manually pre-registering its waker with that `WaitCell` before actually starting the DMA request. This is unfortunate, and it would be nicer if there was an `async fn` for running DMA transfers that the driver could call without having to repeat all of that code.

This branch improves the way we use the D1's DMAC. In particular, rather than hard-coding specific driver-provided `WaitCell`s to correspond to specific DMA channels, we now have the `Dmac` struct track which channels are in use, and allow drivers to allocate any channel from the DMAC. It will provide them with the next available channel if any are not already claimed. The DMAC now owns a `WaitCell` for each DMA channel, which is used to provide an async `Channel::run_descriptor` method. This method handles pre-registering the waker, starting the DMA request, and ensuring the task is woken, and stops the DMA transfer when it's complete.

The allocation of channels by the DMAC means that adding new drivers that also use DMA is easier: we don't need to track who owns which channel. If we don't run a specific driver based on config, any DMA channels it would use are now available to other drivers, rather than being hardcoded. This way, we can have more than 16 driver implementations that use DMA, as long as we don't enable them all at once.

In a follow-up, I'm also going to add dynamic allocation of DMA channels. Currently, DMA channels are still allocated permanently, and drivers allocate them as soon as they're created and hold onto the channel forever. However, we can modify the `Channel` type to also indicate to the DMAC that it's free when the `Channel` is dropped, allowing it to be reallocated (potentially by a different driver). If all 16 DMA channels are in use at the same time, a driver could then _wait_ until one becomes available. If drivers then make more short-lived claims on DMA channels, rather than holding one channel for their entire lifetime, this would thenallow us to have more than 16 driver implementations which use DMA, or allow driver implementations to freely use multiple DMA channels at the same time, without having to worry about running out of channels.

### DIFF
--- a/platforms/allwinner-d1/src/dmac/mod.rs
+++ b/platforms/allwinner-d1/src/dmac/mod.rs
@@ -126,7 +126,7 @@ impl Dmac {
                 }
             }
 
-            // Will write-back and high bits
+            // Will write-back any high bits
             w
         });
 
--- a/platforms/allwinner-d1/src/dmac/mod.rs
+++ b/platforms/allwinner-d1/src/dmac/mod.rs
@@ -138,7 +138,7 @@ impl Dmac {
                 }
             }
 
-            // Will write-back and high bits
+            // Will write-back any high bits
             w
         });
     }
@@ -242,6 +242,20 @@ impl Channel {
         }
     }
 
+    /// # Safety
+    ///
+    /// The caller must ensure that the descriptor pointed to by `desc` is valid
+    /// for as long as the DMA transfer is active. When this function returns,
+    /// the transfer has completed, and it is safe to drop the descriptor.
+    /// However, if this future is cancelled before it completes, the DMA
+    /// transfer may still be active, and must be terminated using [`stop_dma`]
+    /// before the descriptor may be dropped.
+    ///
+    /// # Cancel Safety
+    ///
+    /// Dropping this future does *not* cancel the DMA transfer, and the
+    /// descriptor may not be dropped until the DMA transfer has been canceled.
+    // TODO(eliza): cancel the DMA on drop...
     pub async unsafe fn run_descriptor(&mut self, desc: NonNull<Descriptor>) {
         // mark the channel as in-flight.
         let prev_state = self
@@ -289,6 +303,10 @@ impl Channel {
         })
     }
 
+    /// # Safety
+    ///
+    /// The caller must ensure that the descriptor pointed to by `desc` is valid
+    /// for as long as the DMA transfer is active.
     pub unsafe fn start_descriptor(&mut self, desc: NonNull<Descriptor>) {
         fence(Ordering::SeqCst); //////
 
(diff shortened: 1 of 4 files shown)


platforms/allwinner-d1/src/dmac/mod.rs
```diff
@@ -13,55 +13,162 @@ use d1_pac::{
...
+    channels: [ChannelState; Dmac::CHANNEL_COUNT as usize],
+}
+
+struct ChannelState {
+    waker: WaitCell,
+    state: AtomicU8,
+}
+
+static DMAC_STATE: DmacState = {
+    // This `const` is used as a static initializer, so clippy is wrong here...
```
jamesmunns, line 43: I think we have a few of these, if we want a blanket allow at the top of the project, that's cool with me. It might be across multiple crates sadly tho.

platforms/allwinner-d1/src/dmac/mod.rs
```diff
@@ -13,55 +13,162 @@ use d1_pac::{
...
+        let dmac = unsafe { &*DMAC::PTR };
+        dmac.dmac_irq_pend0.modify(|r, w| {
+            tracing::trace!(dmac_irq_pend0 = ?format_args!("{:#b}", r.bits()), "DMAC interrupt");
+            for i in 0u8..8u8 {
+                if unsafe { r.dma_queue_irq_pend(i) }.bit_is_set() {
+                    DMAC_STATE.channels[i as usize].waker.wake();
+                }
+            }
+
+            // Will write-back and high bits
```
jamesmunns, line 129:
```suggestion
            // Will write-back any high bits
```

platforms/allwinner-d1/src/dmac/mod.rs
```diff
@@ -135,6 +242,39 @@ impl Channel {
...
+        );
+
+        // pre-subscribe to the waitcell to ensure our waker is registered
+        // before starting the DMA transfer.
+        let wait = self.channel.waker.subscribe().await;
+
+        // actually start the DMA transfer.
+        self.start_descriptor(desc);
+
+        // wait for the DMA transfer to complete.
```
jamesmunns, line 264: One thing we still haven't tackled here is that dropping/cancelling/forgetting a DMA future will lead to Very Bad Bonus Memory.

for dropping/cancelling, we could either force-stop the dma transfer (with a manual future that impls Drop), but solving this for `forget` means that the transfer future needs to take owership of the destination buffer, in such a way that forgetting the future also permanently leaks the destination buffer (which might be a whole ass peripheral?).

I'm not against saying "using `forget` in kernel futures is not allowed".

You don't need to fix that issue now, just reminding it exists.

platforms/allwinner-d1/src/dmac/mod.rs
```diff
@@ -177,3 +318,20 @@ pub enum ChannelMode {
     Wait,
     Handshake,
 }
+
```
jamesmunns, line 321: uncomment or remove dead code

jamesmunns (approved):
One big-picture note and a nit or two, but overall looks much better!

platforms/allwinner-d1/src/dmac/mod.rs
hawkw: yeah, i was also thinking about this. i would like to figure out a safe API for it too, but this is why i left the `Channel::run_descriptor` method unsafe for now. we should probably add some docs...

platforms/allwinner-d1/src/dmac/mod.rs
hawkw: unfortunately, `forced_target` means we can't have tests in this crate and run them on the host target. i left this here because i wanted to figure out a solution for that later. i can comment about why it's commented out?

jamesmunns: Sure! that context helps.

platforms/allwinner-d1/src/dmac/mod.rs
jamesmunns: Ah, that's a good point, I think if you mention it in the function's SAFETY docs then we're all good (or at least better off than before!), tho the parent (which might be a safe future) can still be cancelled.

platforms/allwinner-d1/src/dmac/mod.rs
hawkw: i think there might be a clippy config file or something that we could use to disable the lint globally, without having to stick an allow attribute in every crate? it's definitely Not My Favorite Clippy Lint...

platforms/allwinner-d1/src/dmac/mod.rs
hawkw: yeah, we should have some safety docs for this, for sure. i can add them now!

hawkw: left a bunch of comments!

hawkw: we can probably stick something in the `run_descriptor` future that cancels the DMA when it's dropped, making this cancel-safe. but, there's basically no way to make this safe in the face of `mem::forget`, so i think we can just keep the function `unsafe` and document "uwu no forgetty". 

honestly i don't think making driver-facing internal kernel APIs safe with regards to `mem::forget`ting futures is one of my biggest priorities, since it's extremely hard, and if anyone is writing code that's randomly forgetting futures without dropping them, something has gone Really Fucking Wrong somewhere. i would _like_ to be resilient in this case, but i don't think it's always even possible without extreme contortion...

hawkw: at the level of the driver that's running DMA xfers i would expect buffers to be owned either by the driver or some external thing like a shared BBQueue, but inside of the driver i think it's fair to say "driver tasks that `mem::forget` DMA futures deserve whatever happens to them". which, i suppose, means we can't really make these APIs safe. which i'm willing to accept.

PR #281, opened 2023-09-17
feat(D1): DMAC Driver 2: Electric Boogaloo

hawkw:
This branch kinda rewrites a bunch of the DMAC code. Now, DMA channels
are capable of being dynamically allocated on demand using a cool atomic
bitset thingy, and released back to the DMAC pool when they're no longer
in use. This will allow us to have multiple drivers that consume
basically as many DMA channels as they want to, rather than having a
fixed assignment of DMA channels to drivers which has to be maintained
by manually updating the DMAC IRQ handler every time we add a new driver
that wants to do a DMA. A channel can be allocated from the pool once
and used for multiple transfers, if a driver needs to make a bunch of
transfers in sequence and doesn't want to have to round trip through the
channel pool, but they can also be grabbed for oneshot transfers and
immediately released.

I also fixed the cancel safety of the async `Channel::transfer`
(formerly `Channel::run_descriptor`) method, which would previously
leave the transfer running if the future is dropped. This means that if
a caller assumed dropping the transfer future meant that the transfer
descriptor's buffer could be freed, the transfer might complete later
and trample memory that has been reallocated. THAT SEEMS KINDA BAD LOL.
So, now, dropping a `Channel::transfer` future will cancel the transfer
if it hasn't finished yet, allowing fun stuff like attaching a timeout
to a transfer if you really want to. Of course, the transfer may still
have completed partially, and if we were writing to a device, the device
may be unhappy to have only gotten some of the data it wanted. But, at
least we don't have abandoned DMA transfers running around in random
parts of the heap you probably wanted to use for normal stuff like
having strings, or whatever it is that people do on the computer. So
that's better.

Unfortunately, while we can easily make the DMA transfers `Drop`-safe,
we can't really make them `mem::forget`-safe without passing the DMA
buffer into the future and leaking it if it gets forgetted. This is
probably a good idea but would require a nice higher-level interface
around constructing DMA descriptors from buffer-shaped memory regions,
which I didn't want to do tonight. So, the DMA transfer APIs are still
`unsafe` and maybe we can add a nicer thing on top later. In the
meantime, probably don't `mem::forget` DMA transfers. Honestly there is
no reason for normal people to do that so I don't feel super bad about
it.

### DIFF
--- a/platforms/allwinner-d1/src/dmac/mod.rs
+++ b/platforms/allwinner-d1/src/dmac/mod.rs
@@ -330,6 +330,10 @@ impl Dmac {
         });
     }
 
+    /// Cancel *all* currently active DMA transfers.
+    ///
+    /// This is generally used when shutting down the system, such as in panic
+    /// and exception handlers.
     pub(crate) unsafe fn cancel_all() {
         for (i, channel) in DMAC_STATE.channels.iter().enumerate() {
             channel.waker.close();
@@ -463,7 +467,7 @@ impl Channel {
     /// [`Channel::start_descriptor`], it is possible to manipulate the channel
     /// register block while a transfer is in progress. I don't know what
     /// happens if you do this, but it's probably bad.
-    pub unsafe fn desc_addr_reg(&self) -> &Reg<DMAC_DESC_ADDR_SPEC> {
+    unsafe fn desc_addr_reg(&self) -> &Reg<DMAC_DESC_ADDR_SPEC> {
         let dmac = &*DMAC::PTR;
         match self.idx {
             0 => &dmac.dmac_desc_addr0,
@@ -505,7 +509,7 @@ impl Channel {
     /// [`Channel::start_descriptor`], it is possible to manipulate the channel
     /// register block while a transfer is in progress. I don't know what
     /// happens if you do this, but it's probably bad.
-    pub unsafe fn en_reg(&self) -> &Reg<DMAC_EN_SPEC> {
+    unsafe fn en_reg(&self) -> &Reg<DMAC_EN_SPEC> {
         let dmac = &*DMAC::PTR;
         match self.idx {
             0 => &dmac.dmac_en0,
@@ -547,7 +551,7 @@ impl Channel {
     /// [`Channel::start_descriptor`], it is possible to manipulate the channel
     /// register block while a transfer is in progress. I don't know what
     /// happens if you do this, but it's probably bad.
-    pub unsafe fn mode_reg(&self) -> &Reg<DMAC_MODE_SPEC> {
+    unsafe fn mode_reg(&self) -> &Reg<DMAC_MODE_SPEC> {
         let dmac = &*DMAC::PTR;
         match self.idx {
             0 => &dmac.dmac_mode0,
@@ -586,7 +590,7 @@ impl Channel {
     /// The caller must not initiate another transfer on this channel until the
     /// transfer started using `start_descriptor` completes. I don't know what
     /// happens if you do this, but I'm sure it's bad.
-    pub unsafe fn start_descriptor(&mut self, desc: NonNull<Descriptor>) {
+    unsafe fn start_descriptor(&mut self, desc: NonNull<Descriptor>) {
         fence(Ordering::SeqCst); //////
 
         let desc_addr = desc.as_ptr() as usize;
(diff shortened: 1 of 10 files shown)


platforms/allwinner-d1/src/dmac/mod.rs
```diff
@@ -30,115 +32,198 @@ pub struct Channel {
...
-            state: &DMAC_STATE,
-            dmac,
+        /// Sets the `DMA_QUEUE_IRQ_EN` bit for the given channel index.
+        fn set_queue_irq_en(idx: u8, bits: u32) -> u32 {
+            bits | (1 << queue_irq_en_offset(idx))
         }
+
+        ccu.enable_module(&mut dmac);
+
+        // enable the queue IRQ for all the channels, because use the other
```
hawkw, line 70:
```suggestion
        // enable the queue IRQ for all the channels, because we don't use the other
```

platforms/allwinner-d1/src/dmac/mod.rs
```diff
@@ -13,167 +13,481 @@ use d1_pac::{
...
+/// source and destination modes of a [`Channel`] may be configured using
+/// [`Channel::set_channel_modes`].
+#[derive(Copy, Clone, Debug)]
+pub enum ChannelMode {
+    /// DMA transfer wait mode.
+    ///
+    /// In this mode, the DMAC will wait for a configurable number of clock
+    /// cycles before automatically starting the next transfer.
+    ///
+    /// The Allwinner documentationh for the D1 describes this mode as follows:
```
jamesmunns, line 74:
```suggestion
    /// The Allwinner documentation for the D1 describes this mode as follows:
```

platforms/allwinner-d1/src/dmac/mod.rs
```diff
@@ -13,167 +13,481 @@ use d1_pac::{
...
 pub struct Channel {
     idx: u8,
-    channel: &'static ChannelState,
+    xfer_done: &'static WaitCell,
 }
 
-struct DmacState {
-    used_channels: AtomicU8,
-    channels: [ChannelState; Dmac::CHANNEL_COUNT as usize],
+/// DMA channel modes.
```
jamesmunns, line 62: love love love these docs.

platforms/allwinner-d1/src/dmac/mod.rs
```diff
@@ -13,167 +13,481 @@ use d1_pac::{
...
+    /// Of course, the transfer may still have completed partially, and if we
+    /// were writing to a device, the device may be unhappy to have only gotten
+    /// some of the data it wanted. Cancelling an incomplete transfer may result
+    /// in, for example, writing out half of a string to the UART, or only part
+    /// of a structured message over SPI, and so on. But, at least we don't have
+    /// abandoned DMA transfers running around in random parts of the heap you
+    /// probably wanted to use for normal stuff like having strings, or whatever
+    /// it is that people do on the computer.
+    ///
+    /// If the DMA transfer was a read rather than a write, cancelling a partial
```
jamesmunns, line 229: I'd say this isn't TOTALLY true, depending on how broad you are talking about, reads can still have side effects in external hardware.

source/bitslab/src/index.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// Release an index back to the pool.
+    ///
+    /// The freed index may now be returned by a subsequent call to [`allocate`](Self::allocate).
+    #[inline]
+    pub fn free(&self, index: u8) {
+        self.0.fetch_and(!(1 << index), Release);
+    }
+
+    /// Returns `true` if all indices in the allocator have been allocated.
```
jamesmunns, line 41: are the docs for full/empty backwards?

jamesmunns: no wait, this is just a reverse perspective thing, "full => no allocation"/"empty => all allocated". *might* be worth using something other than "full"/"empty", like `all_available`/`none_available`, but maybe I'm just silly.

jamesmunns (approved):
This is hella hella cool.

I also really like that you can still reserve and take a DMA channel and re-use it, which addressed my one potential concern of "how do we handle 'high prio' dma cases"?

Looks like we're all the way ready for `async fn copy(src: &[u8], dst: &mut [u8])` now :)

platforms/allwinner-d1/src/dmac/mod.rs
hawkw: yeah, good point

source/bitslab/src/index.rs
hawkw: yeah, i went back and forth on what full and empty meant in this context. i think maybe i should change them to something more like what you suggested --- good call!

PR #284, opened 2023-09-17
feat(D1): add builders for DMAC descriptors

hawkw:
Currently, the DMAC descriptor configuration fields are bit-packed every
time a descriptor is constructed. However, the drivers which use DMA
always construct descriptors with the same configuration bitfields --
only the addresses and lengths change. We can make this more efficient
by constructing the config field once and copying it into each new
descriptor.

This branch replaces the existing `DescriptorConfig` struct with a
`DescriptorBuilder`, which is basically the same thing except it
actually constructs the bitfield as a single u32. I used
`mycelium-bitfield` to do this --- mostly for fun, but it's maybe a bit
more readable than the previous code? I dunno.

### DIFF
--- a/platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs
+++ b/platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs
@@ -52,9 +52,12 @@ pub enum InvalidOperandReason {
     /// memory. Operand addresses for DMA transfers may not exceed
     /// [`Descriptor::ADDR_MAX`] (34 bits).
     AddrTooHigh(usize),
+    /// A slice was longer than the maximum supported DMA transfer size of
+    /// [`Descriptor::MAX_LEN`] (25 bits).
+    TooLong(usize),
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum Operand {
     Source,
     Destination,
@@ -285,15 +288,22 @@ impl<S, D> DescriptorBuilder<S, D> {
     ///
     /// # Returns
     ///
-    /// - [`Ok`]`(`[`DescriptorBuilder`]`)` with `source` as the source operand,
-    ///   if the provided slice's address is a valid DMA source.
-    /// - [`Err`]`(`[`InvalidOperand`]`)` if `source` is not a valid DMA
-    ///   source address.
+    /// - [`Ok`]`(`[`DescriptorBuilder`]`)` with `dest` as the destination
+    ///   operand, if the provided slice's address is a valid DMA destination.
+    /// - [`Err`]`(`[`InvalidOperand`]`)` with [`InvalidOperandReason::TooLong`]
+    ///   if the provided slice is longer than [`Descriptor::MAX_LEN`].
+    /// - [`Err`]`(`[`InvalidOperand`]`)` with
+    ///   [`InvalidOperandReason::AddrTooHigh`] if the provided slice's address
+    ///   is higher than [`Descriptor::ADDR_MAX`].
     pub fn source_slice(
         self,
         source: &'_ [u8],
     ) -> Result<DescriptorBuilder<&'_ [u8], D>, InvalidOperand> {
         Self::check_addr(source as *const _ as *const (), Operand::Source)?;
+
+        if source.len() > Descriptor::MAX_LEN as usize {
+            return Err(InvalidOperand::too_long(Operand::Source, source.len()));
+        }
 
         Ok(DescriptorBuilder {
             cfg: self
@@ -311,19 +321,27 @@ impl<S, D> DescriptorBuilder<S, D> {
     /// be copied from the source operand of the transfer into this slice.
     ///
     /// Since the slice is in memory, this automatically sets the destination address
-    /// mode to [`AddressMode::LinearMode`].
+    /// mode to [`AddressMode::LinearMode`], and the destination DRQ type to
+    /// [`DestDrqType::Dram].
     ///
     /// # Returns
     ///
     /// - [`Ok`]`(`[`DescriptorBuilder`]`)` with `dest` as the destination
     ///   operand, if the provided slice's address is a valid DMA destination.
-    /// - [`Err`]`(`[`InvalidOperand`]`)` if `dest` is not a valid DMA
-    ///   destination address.
+    /// - [`Err`]`(`[`InvalidOperand`]`)` with [`InvalidOperandReason::TooLong`]
+    ///   if the provided slice is longer than [`Descriptor::MAX_LEN`].
+    /// - [`Err`]`(`[`InvalidOperand`]`)` with
+    ///   [`InvalidOperandReason::AddrTooHigh`] if the provided slice's address
+    ///   is higher than [`Descriptor::ADDR_MAX`].
     pub fn dest_slice(
         self,
         dest: DestBuf<'_>,
     ) -> Result<DescriptorBuilder<S, DestBuf<'_>>, InvalidOperand> {
         Self::check_addr(dest as *const _ as *const (), Operand::Destination)?;
+
+        if dest.len() > Descriptor::MAX_LEN as usize {
+            return Err(InvalidOperand::too_long(Operand::Destination, dest.len()));
+        }
 
         Ok(DescriptorBuilder {
             cfg: self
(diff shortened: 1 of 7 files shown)


platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs
```diff
@@ -8,299 +8,389 @@
...
+        }
+        if destination >= (1 << 34) {
+            return Err("destination address may not exceed 34 bits");
+        }
+        if byte_counter >= (1 << 25) {
+            return Err("byte counter may not exceed 25 bits");
+        }
+        if let Some(link) = self.link {
+            if (link as usize & 0b11) != 0 {
+                return Err("link address's lowest two bits must be unset");
```
jamesmunns, line 272:
```suggestion
                return Err("linked descriptors must be at least 4-byte aligned");
```

platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs
```diff
@@ -8,299 +8,389 @@
...
     }
-}
 
-// BModeSel
-
-impl BModeSel {
-    #[inline(always)]
-    fn to_desc_bits(self) -> u32 {
-        (((self as u8) & 0b1) as u32) << 30
+    proptest! {
```
jamesmunns, line 373: *helllllll yeah*

platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs
```diff
@@ -8,299 +8,389 @@
...
+            wait_clock_cycles,
+            ..self
         }
-
-        Ok(descriptor)
     }
-}
 
-// DescriptorConfig
+    pub fn build(
```
jamesmunns, line 252: would it make sense to have an outer method that is generic over `<S, D>` types so the caller doesn't have to do pointer shenanigans?

so rename THIS function as `build_inner`, then something like:

```rust
fn build<S, D>(self, src: &S, dst: &mut D, byte_counter: u32) -> Result<...> {
    let src = src.as_ptr().cast::<()>();
    let dst = dst.as_mut_ptr().cast::<()>();
    build_inner(src, dst, byte_counter)
}
```

just as a little QoL thing.

jamesmunns (approved):
This is super cool, thank you!

platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs
hawkw: yeah that's nicer!

platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs
hawkw: hmm, that's not a bad idea, we could do something like that. potentially we could even have a version where the destination is taken as a `&mut [MaybeUninit<u8>]` and the length is used to generate `byte_counter`...

hawkw: i wonder if `dst` in this should actually be `Pin<&mut D>` so that we can prove we're not going to move the pointed buffer out from under the caller? maybe that's not worth it though idk.

hawkw: hmm, i tried adding a function like this, but i'm not sure if it's worth it, since it actually makes both existing uses of `build` _more_ complex, as we need to turn the raw pointers to the various MMIO dest registers into `&mut whatever`s, just so we can pass them back to the function that turns them into raw pointers:

```patch
diff --git a/platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs b/platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs
index 9090920..583456f 100644
--- a/platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs
+++ b/platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs
@@ -272,7 +272,18 @@ impl DescriptorBuilder {
         }
     }
 
-    pub fn build(
+    pub fn build<S, D>(
+        self,
+        source: &S,
+        destination: &mut D,
+        byte_counter: u32,
+    ) -> Result<Descriptor, InvalidDescriptor> {
+        let source = source as *const _ as *const ();
+        let destination = destination as *mut _ as *mut ();
+        self.build_raw(source, destination, byte_counter)
+    }
+
+    pub fn build_raw(
         self,
         source: *const (),
         destination: *mut (),
diff --git a/platforms/allwinner-d1/d1-core/src/drivers/spim.rs b/platforms/allwinner-d1/d1-core/src/drivers/spim.rs
index c853f43..9806a0a 100644
--- a/platforms/allwinner-d1/d1-core/src/drivers/spim.rs
+++ b/platforms/allwinner-d1/d1-core/src/drivers/spim.rs
@@ -118,9 +118,7 @@ impl SpiSenderServer {
             .spawn(async move {
                 let spi = unsafe { &*SPI_DBI::PTR };
 
-                let txd_ptr: *mut u32 = spi.spi_txd.as_ptr();
-                let txd_ptr: *mut u8 = txd_ptr.cast();
-                let txd_ptr: *mut () = txd_ptr.cast();
+                let txd = unsafe { &mut *(spi.spi_txd.as_ptr() as *mut _) };
 
                 let descr_cfg = Descriptor::builder()
                     .dest_data_width(DataWidth::Bit8)
@@ -137,20 +135,23 @@ impl SpiSenderServer {
                     let Message { msg, reply } = reqs.next_request().await;
                     let SpiSenderRequest::Send(ref payload) = msg.body;
 
-                    let len = payload.as_slice().len();
+                    let payload = payload.as_slice();
+                    // TODO(eliza): we should probably respond with an error if
+                    // someone tries to send a payload whose length is > u32::MAX...;
+                    let len = payload.len() as u32;
 
                     spi.spi_bcc.modify(|_r, w| {
                         // "Single Mode Transmit Counter" - the number of bytes to send
-                        w.stc().variant(len as u32);
+                        w.stc().variant(len);
                         w
                     });
                     spi.spi_mbc.modify(|_r, w| {
                         // Master Burst Counter
-                        w.mbc().variant(len as u32);
+                        w.mbc().variant(len);
                         w
                     });
                     spi.spi_mtc.modify(|_r, w| {
-                        w.mwtc().variant(len as u32);
+                        w.mwtc().variant(len);
                         w
                     });
                     // Start transfer
@@ -160,7 +161,7 @@ impl SpiSenderServer {
                     });
 
                     let descriptor = descr_cfg
-                        .build(payload.as_slice().as_ptr().cast(), txd_ptr, len as u32)
+                        .build(&payload[0], txd, len)
                         .expect("failed to build SPI1_TX DMA descriptor");
 
                     // start the DMA transfer.
diff --git a/platforms/allwinner-d1/d1-core/src/drivers/uart.rs b/platforms/allwinner-d1/d1-core/src/drivers/uart.rs
index 2b9ba2f..db9ad9e 100644
--- a/platforms/allwinner-d1/d1-core/src/drivers/uart.rs
+++ b/platforms/allwinner-d1/d1-core/src/drivers/uart.rs
@@ -124,12 +124,12 @@ impl D1Uart {
         loop {
             let rx = cons.read_grant().await;
             let len = rx.len();
-            let thr_addr = unsafe { &*UART0::PTR }.thr() as *const _ as *mut ();
+            let thr = unsafe { &mut *((*UART0::PTR).thr() as *const _ as *mut u32) };
 
             let rx_sli: &[u8] = &rx;
 
             let descriptor = descr_cfg
-                .build(rx_sli.as_ptr().cast(), thr_addr, rx_sli.len() as u32)
+                .build(&rx_sli[0], thr, rx_sli.len() as u32)
                 .expect("failed to build UART0 DMA transfer descriptor");
 
             // start the DMA transfer.

```
wdyt?

hawkw: ...aaand `rustc` actually does not like that code at all:
```
error: casting `&T` to `&mut T` is undefined behavior, even if the reference is unused, consider instead using an `UnsafeCell`
   --> platforms/allwinner-d1/d1-core/src/drivers/uart.rs:127:32
    |
127 |             let thr = unsafe { &mut *((*UART0::PTR).thr() as *const _ as *mut u32) };
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(invalid_reference_casting)]` on by default
```
so we probably need to actually just use the raw-ptr `build_raw` regardless...

hawkw:
hmm, apparently something on this branch is actually making display updates way laggier than on `main`. gotta figure out what i did wrong!

platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs
hawkw: as of 42671b95637b383315f8933ccc2012ed0158a82a, i've figured out a nice way to do this. :)

PR #299, opened 2023-10-28
feat(d1): add smhc driver

jspngh:
This PR adds support for SD card interaction on the Allwinner D1 platform.
It is split in an `sdmmc` kernel service and a driver for the SD/MMC host controller peripheral.
Currently only SD card support is implemented, but this should allow for adding MMC and SDIO support later.
The interface between the kernel service and the driver aims to work on the specification/protocol level
and not contain any Allwinner D1 details, which should make it suitable for other platforms as well.

This PR also improves the `BusGatingResetRegister` trait in `platforms/allwinner-d1/d1-core/src/ccu.rs`

### DIFF
--- a/platforms/allwinner-d1/d1-core/src/ccu.rs
+++ b/platforms/allwinner-d1/d1-core/src/ccu.rs
@@ -20,10 +20,10 @@ pub struct Ccu {
 
 /// Trait to be implemented for module clocks that can be gated and reset
 pub trait BusGatingResetRegister {
-    /// Enable or disable the clock reset bit
-    fn gating(ccu: &mut CCU, pass: bool);
     /// Enable or disable the clock gating bit
-    fn reset(ccu: &mut CCU, deassert: bool);
+    fn gating(ccu: &mut CCU, pass: bool);
+    /// Enable or disable the clock reset bit
+    fn reset(ccu: &mut CCU, assert: bool);
 }
 
 // TODO: should this move into the `Clint`?
@@ -49,7 +49,7 @@ impl Ccu {
 
     /// De-assert the reset bit and enable the clock gating bit for the given module
     pub fn enable_module<MODULE: BusGatingResetRegister>(&mut self, _mod: &mut MODULE) {
-        MODULE::reset(&mut self.ccu, true);
+        MODULE::reset(&mut self.ccu, false);
         sdelay(20);
         MODULE::gating(&mut self.ccu, true);
     }
@@ -58,7 +58,7 @@ impl Ccu {
     pub fn disable_module<MODULE: BusGatingResetRegister>(&mut self, _mod: &mut MODULE) {
         MODULE::gating(&mut self.ccu, false);
         // TODO: delay?
-        MODULE::reset(&mut self.ccu, false);
+        MODULE::reset(&mut self.ccu, true);
     }
 
     /// Allow modules to configure their own clock on a PAC level
@@ -252,8 +252,8 @@ macro_rules! impl_bgr {
                     ccu.$reg.modify(|_, w| w.$gating().bit(pass));
                 }
 
-                fn reset(ccu: &mut CCU, deassert: bool) {
-                    ccu.$reg.modify(|_, w| w.$reset().bit(deassert));
+                fn reset(ccu: &mut CCU, assert: bool) {
+                    ccu.$reg.modify(|_, w| w.$reset().bit(!assert));
                 }
             }
         )+
--- a/platforms/allwinner-d1/d1-core/src/ccu.rs
+++ b/platforms/allwinner-d1/d1-core/src/ccu.rs
@@ -262,21 +262,13 @@ macro_rules! impl_bgr {
             impl BusGatingResetRegister for $MODULE {
                 fn gating(ccu: &mut CCU, gating: BusGating) {
                     ccu.$reg.modify(|_, w| {
-                        if gating == BusGating::Mask {
-                            w.$gating().clear_bit()
-                        } else {
-                            w.$gating().set_bit()
-                        }
+                        w.$gating().bit(gating == BusGating::Pass)
                     });
                 }
 
                 fn reset(ccu: &mut CCU, reset: BusReset) {
                     ccu.$reg.modify(|_, w| {
-                        if reset == BusReset::Assert {
-                            w.$reset().clear_bit()
-                        } else {
-                            w.$reset().set_bit()
-                        }
+                        w.$reset().bit(reset == BusReset::Deassert)
                     });
                 }
             }
--- a/platforms/allwinner-d1/d1-core/src/ccu.rs
+++ b/platforms/allwinner-d1/d1-core/src/ccu.rs
@@ -18,12 +18,24 @@ pub struct Ccu {
     ccu: CCU,
 }
 
+#[derive(PartialEq)]
+pub enum BusGating {
+    Mask,
+    Pass,
+}
+
+#[derive(PartialEq)]
+pub enum BusReset {
+    Assert,
+    Deassert,
+}
+
 /// Trait to be implemented for module clocks that can be gated and reset
 pub trait BusGatingResetRegister {
     /// Enable or disable the clock gating bit
-    fn gating(ccu: &mut CCU, pass: bool);
+    fn gating(ccu: &mut CCU, gating: BusGating);
     /// Enable or disable the clock reset bit
-    fn reset(ccu: &mut CCU, assert: bool);
+    fn reset(ccu: &mut CCU, reset: BusReset);
 }
 
 // TODO: should this move into the `Clint`?
@@ -49,16 +61,16 @@ impl Ccu {
 
     /// De-assert the reset bit and enable the clock gating bit for the given module
     pub fn enable_module<MODULE: BusGatingResetRegister>(&mut self, _mod: &mut MODULE) {
-        MODULE::reset(&mut self.ccu, false);
+        MODULE::reset(&mut self.ccu, BusReset::Deassert);
         sdelay(20);
-        MODULE::gating(&mut self.ccu, true);
+        MODULE::gating(&mut self.ccu, BusGating::Pass);
     }
 
     /// Disable the clock gating bit and assert the reset bit for the given module
     pub fn disable_module<MODULE: BusGatingResetRegister>(&mut self, _mod: &mut MODULE) {
-        MODULE::gating(&mut self.ccu, false);
+        MODULE::gating(&mut self.ccu, BusGating::Mask);
         // TODO: delay?
-        MODULE::reset(&mut self.ccu, true);
+        MODULE::reset(&mut self.ccu, BusReset::Assert);
     }
 
     /// Allow modules to configure their own clock on a PAC level
@@ -248,12 +260,24 @@ macro_rules! impl_bgr {
     ($($MODULE:ident : ($reg:ident, $gating:ident, $reset:ident),)+) => {
         $(
             impl BusGatingResetRegister for $MODULE {
-                fn gating(ccu: &mut CCU, pass: bool) {
-                    ccu.$reg.modify(|_, w| w.$gating().bit(pass));
+                fn gating(ccu: &mut CCU, gating: BusGating) {
+                    ccu.$reg.modify(|_, w| {
+                        if gating == BusGating::Mask {
+                            w.$gating().clear_bit()
+                        } else {
+                            w.$gating().set_bit()
+                        }
+                    });
                 }
 
-                fn reset(ccu: &mut CCU, assert: bool) {
-                    ccu.$reg.modify(|_, w| w.$reset().bit(!assert));
+                fn reset(ccu: &mut CCU, reset: BusReset) {
+                    ccu.$reg.modify(|_, w| {
+                        if reset == BusReset::Assert {
+                            w.$reset().clear_bit()
+                        } else {
+                            w.$reset().set_bit()
+                        }
+                    });
                 }
             }
         )+
--- a/platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
+++ b/platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
@@ -511,12 +511,13 @@ impl Smhc {
                 ErrorKind::Other => sdmmc::Error::Other,
             })
         } else if long_resp {
-            Ok(sdmmc::Response::Long([
+            let rsp: [u32; 4] = [
                 self.smhc.smhc_resp0.read().bits(),
                 self.smhc.smhc_resp1.read().bits(),
                 self.smhc.smhc_resp2.read().bits(),
                 self.smhc.smhc_resp3.read().bits(),
-            ]))
+            ];
+            Ok(sdmmc::Response::Long(unsafe { core::mem::transmute(rsp) }))
         } else {
             Ok(sdmmc::Response::Short {
                 value: self.smhc.smhc_resp0.read().bits(),
--- a/platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
+++ b/platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
@@ -296,6 +296,18 @@ impl Smhc {
         let long_resp = params.rsp_type == sdmmc::ResponseType::Long;
         let resp_rcv = params.rsp_type != sdmmc::ResponseType::None;
 
+        // Do any required configuration
+        match params.options {
+            sdmmc::HardwareOptions::None => (),
+            sdmmc::HardwareOptions::SetBusWidth(width) => {
+                self.smhc.smhc_ctype.write(|w| match width {
+                    sdmmc::BusWidth::Single => w.card_wid().b1(),
+                    sdmmc::BusWidth::Quad => w.card_wid().b4(),
+                    sdmmc::BusWidth::Octo => w.card_wid().b8(),
+                });
+            }
+        }
+
         // Set the argument to be sent
         self.smhc
             .smhc_cmdarg
--- a/platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
+++ b/platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
@@ -141,11 +141,11 @@ impl Smhc {
         smhc.smhc_clkdiv.write(|w| w.cclk_enb().off());
 
         ccu.disable_module(&mut smhc);
-        // Set module clock rate to 200 MHz
+        // Set module clock rate to 100 MHz
         // TODO: ccu should provide a higher-level abstraction for this
         ccu.borrow_raw().smhc0_clk.write(|w| {
             w.clk_src_sel().pll_peri_1x();
-            w.factor_n().variant(d1_pac::ccu::smhc0_clk::FACTOR_N_A::N1);
+            w.factor_n().variant(d1_pac::ccu::smhc0_clk::FACTOR_N_A::N2);
             w.factor_m().variant(2);
             w.clk_gating().set_bit();
             w
--- a/platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
+++ b/platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
@@ -16,7 +16,6 @@ use core::{
 use crate::ccu::{BusGatingResetRegister, Ccu};
 use d1_pac::{smhc, Interrupt, GPIO, SMHC0, SMHC1, SMHC2};
 use kernel::{
-    comms::kchannel::{KChannel, KConsumer},
     mnemos_alloc::containers::FixedVec,
     registry,
     services::sdmmc::{self, SdmmcService},
@@ -62,12 +61,12 @@ enum SmhcOp {
     Control,
     Read {
         buf: FixedVec<u8>,
-        cnt: u32,
+        cnt: usize,
         auto_stop: bool,
     },
     Write {
         buf: FixedVec<u8>,
-        cnt: u32,
+        cnt: usize,
         auto_stop: bool,
     },
 }
@@ -79,6 +78,8 @@ enum State {
     Idle,
     /// Waiting for command to be completed.
     WaitForCommand,
+    /// Waiting for the DMA operation to be completed.
+    WaitForDma,
     /// Waiting for data transfer to be completed.
     WaitForDataTransfer,
     /// Waiting for the auto-stop command to be sent and completed.
@@ -167,6 +168,13 @@ impl Smhc {
             w.re_int_en().set_bit();
             w
         });
+        smhc.smhc_idie.write(|w| {
+            w.des_unavl_int_enb().set_bit();
+            w.ferr_int_enb().set_bit();
+            w.rx_int_enb().set_bit();
+            w.tx_int_enb().set_bit();
+            w
+        });
 
         Self::init(
             unsafe { &*SMHC0::ptr() },
@@ -216,7 +224,8 @@ impl Smhc {
         // Set the sample delay to 0 (as done in Linux and Allwinner BSP)
         smhc.smhc_smap_dl.write(|w| {
             w.samp_dl_sw().variant(0);
-            w.samp_dl_sw_en().set_bit()
+            w.samp_dl_sw_en().set_bit();
+            w
         });
 
         // Enable the card clock
@@ -233,6 +242,60 @@ impl Smhc {
             smhc,
             isr: &SMHC0_ISR,
             int: (int, isr),
+        }
+    }
+
+    /// # Safety
+    /// The descriptor chain needs to live at least as long as the DMA transfer.
+    /// Additionally, their content (e.g., the validity of the buffers they point to)
+    /// also needs to be verified by the user.
+    unsafe fn prepare_dma(&self, descriptor: &idmac::Descriptor, byte_cnt: u32) {
+        self.smhc.smhc_ctrl.modify(|_, w| {
+            w.dma_enb().set_bit();
+            w.dma_rst().set_bit();
+            w
+        });
+        while self.smhc.smhc_ctrl.read().dma_rst().bit_is_set() {}
+
+        // Configure the address of the first DMA descriptor
+        // Right-shift by 2 because it is a *word-address*.
+        self.smhc
+            .smhc_dlba
+            .write(|w| unsafe { w.bits((descriptor as *const _ as u32) >> 2) });
+
+        // Set number of bytes that will be read or written.
+        self.smhc.smhc_bytcnt.write(|w| unsafe { w.bits(byte_cnt) });
+
+        // Soft reset of DMA controller
+        self.smhc.smhc_idmac.write(|w| w.idmac_rst().set_bit());
+
+        // Configure the burst size and TX/RX trigger level
+        self.smhc.smhc_fifoth.write(|w| {
+            w.tx_tl().variant(8);
+            w.rx_tl().variant(7);
+            w.bsize_of_trans().t8();
+            w
+        });
+
+        // Configure the transfer interrupt, receive interrupt, and abnormal interrupt.
+        self.smhc.smhc_idie.write(|w| {
+            w.rx_int_enb().set_bit();
+            w.tx_int_enb().set_bit();
+            w.err_sum_int_enb().set_bit();
+            w
+        });
+
+        // Enable the IDMAC and configure burst transfers
+        self.smhc.smhc_idmac.write(|w| {
+            w.idmac_enb().set_bit();
+            w.fix_bust_ctrl().set_bit();
+            w
+        });
+
+        // Reset the FIFO
+        self.smhc.smhc_ctrl.modify(|_, w| w.fifo_rst().reset());
+        while self.smhc.smhc_ctrl.read().fifo_rst().is_reset() {
+            core::hint::spin_loop();
         }
     }
 
@@ -308,6 +371,75 @@ impl Smhc {
             }
         }
 
+        let mut guard = self.isr.lock(self.smhc);
+        guard.data.op = match (params.kind, params.buffer) {
+            (sdmmc::CommandKind::Control, _) => SmhcOp::Control,
+            (sdmmc::CommandKind::Read(cnt), Some(buf)) => SmhcOp::Read {
+                buf,
+                cnt,
+                auto_stop,
+            },
+            (sdmmc::CommandKind::Write(cnt), Some(buf)) => SmhcOp::Write {
+                buf,
+                cnt,
+                auto_stop,
+            },
+            _ => {
+                tracing::error!("did not provide a buffer for read/write");
+                return Err(sdmmc::Error::Other);
+            }
+        };
+
+        // Statically allocate space for 16 DMA descriptors.
+        // Each descriptor can do a transfer of 4KB, giving a max total transfer of 64KB.
+        // By declaring them in this scope the descriptor memory will live long enough,
+        // however currently this is up to the user to guarantuee.
+        // Safety: a zero'ed descriptor is valid and will simply be ignored by the IDMAC.
+        let mut descriptors: [idmac::Descriptor; 16] = unsafe { core::mem::zeroed() };
+
+        // Perform checks on arguments to make sure we won't overflow the buffer
+        match &guard.data.op {
+            SmhcOp::Read { buf, cnt, .. } | SmhcOp::Write { buf, cnt, .. } => {
+                // Currently we limit the number of data that can be read at once
+                if *cnt > 0x10000 || buf.capacity() < *cnt {
+                    return Err(sdmmc::Error::Data);
+                }
+
+                tracing::debug!("Creating descriptor chain from buffer");
+                let buf_ptr = buf.as_slice().as_ptr();
+                let mut remaining = *cnt;
+                let mut index = 0;
+                while remaining > 0x1000 {
+                    descriptors[index] = idmac::Descriptor::try_from(idmac::DescriptorConfig {
+                        disable_int_on_complete: true,
+                        first: index == 0,
+                        last: false,
+                        buff_size: 0x1000,
+                        buff_addr: unsafe { buf_ptr.add(index * 0x1000) },
+                        next_desc: Some(&descriptors[index + 1] as *const _ as _),
+                    })
+                    .map_err(|_| sdmmc::Error::Other)?;
+                    remaining -= 0x1000;
+                    index += 1;
+                }
+
+                descriptors[index] = idmac::Descriptor::try_from(idmac::DescriptorConfig {
+                    disable_int_on_complete: false,
+                    first: index == 0,
+                    last: true,
+                    buff_size: remaining as u16,
+                    buff_addr: unsafe { buf_ptr.add(index * 0x1000) },
+                    next_desc: None,
+                })
+                .map_err(|_| sdmmc::Error::Other)?;
+
+                // TODO: should this return some kind of guard, which needs to be used later
+                // to make sure the descriptor has a long enough lifetime?
+                unsafe { self.prepare_dma(&descriptors[0], *cnt as u32) };
+            }
+            _ => (),
+        }
+
         // Set the argument to be sent
         self.smhc
             .smhc_cmdarg
@@ -327,27 +459,8 @@ impl Smhc {
             w
         });
 
-        // Now wait for completion or error interrupt
-        let mut guard = self.isr.lock(self.smhc);
         guard.data.state = State::WaitForCommand;
-        guard.data.op = match (params.kind, params.buffer) {
-            (sdmmc::CommandKind::Control, _) => SmhcOp::Control,
-            (sdmmc::CommandKind::Read(cnt), Some(buf)) => SmhcOp::Read {
-                buf,
-                cnt,
-                auto_stop,
-            },
-            (sdmmc::CommandKind::Write(cnt), Some(buf)) => SmhcOp::Write {
-                buf,
-                cnt,
-                auto_stop,
-            },
-            _ => {
-                tracing::warn!("did not provide a buffer for read/write");
-                return Err(sdmmc::Error::Other);
-            }
-        };
-
+        // Now wait for completion or error interrupt
         guard.wait_for_irq().await;
         tracing::trace!("SMHC operation completed");
         let res = if let Some(error) = guard.data.err.take() {
@@ -425,70 +538,103 @@ impl DerefMut for SmhcDataGuard<'_> {
 }
 
 impl SmhcData {
+    /// Check for error bits in the interrupt status registers.
     fn error_status(smhc: &smhc::RegisterBlock) -> Option<ErrorKind> {
         let mint = smhc.smhc_mintsts.read();
-        if mint.m_rto_back_int().bit_is_set() {
-            return Some(ErrorKind::ResponseTimeout);
-        }
-        if mint.m_rce_int().bit_is_set() {
-            return Some(ErrorKind::ResponseCrc);
-        }
-        if mint.m_re_int().bit_is_set() {
-            return Some(ErrorKind::Response);
-        }
-        if mint.m_dsto_vsd_int().bit_is_set() {
-            return Some(ErrorKind::DataStarvationTimeout);
-        }
-        if mint.m_dto_bds_int().bit_is_set() {
-            return Some(ErrorKind::DataTimeout);
-        }
-        if mint.m_dce_int().bit_is_set() {
-            return Some(ErrorKind::DataCrc);
-        }
-        if mint.m_fu_fo_int().bit_is_set() {
-            return Some(ErrorKind::FifoUnderrunOverflow);
-        }
-        if mint.m_cb_iw_int().bit_is_set() {
-            return Some(ErrorKind::CommandBusyIllegalWrite);
-        }
-        if mint.m_dse_bc_int().bit_is_set() {
-            return Some(ErrorKind::DataStart);
-        }
-        if mint.m_dee_int().bit_is_set() {
-            return Some(ErrorKind::DataEnd);
-        }
+        let idst = smhc.smhc_idst.read();
 
-        None
+        if mint.m_dsto_vsd_int().bit_is_set() {
+            Some(ErrorKind::DataStarvationTimeout)
+        } else if mint.m_fu_fo_int().bit_is_set() {
+            Some(ErrorKind::FifoUnderrunOverflow)
+        } else if mint.m_cb_iw_int().bit_is_set() {
+            Some(ErrorKind::CommandBusyIllegalWrite)
+        } else if mint.m_rto_back_int().bit_is_set() {
+            Some(ErrorKind::ResponseTimeout)
+        } else if mint.m_rce_int().bit_is_set() {
+            Some(ErrorKind::ResponseCrc)
+        } else if mint.m_re_int().bit_is_set() {
+            Some(ErrorKind::Response)
+        } else if mint.m_dto_bds_int().bit_is_set() {
+            Some(ErrorKind::DataTimeout)
+        } else if mint.m_dce_int().bit_is_set() {
+            Some(ErrorKind::DataCrc)
+        } else if mint.m_dse_bc_int().bit_is_set() {
+            Some(ErrorKind::DataStart)
+        } else if mint.m_dee_int().bit_is_set() {
+            Some(ErrorKind::DataEnd)
+        } else if idst.des_unavl_int().bit_is_set() || idst.fatal_berr_int().bit_is_set() {
+            Some(ErrorKind::Dma)
+        } else {
+            None
+        }
     }
 
     fn advance_isr(&mut self, smhc: &smhc::RegisterBlock, num: u8) {
         tracing::trace!(state = ?self.state, smhc = num, "SMHC{num} interrupt");
 
+        let mut needs_wake = false;
+
         self.err = Self::error_status(smhc);
-        let mut needs_wake = self.err.is_some();
+        if let Some(err) = self.err {
+            // Clear all interrupt bits
+            smhc.smhc_rintsts.write(|w| unsafe { w.bits(0xFFFF_FFFF) });
+            smhc.smhc_idst.write(|w| unsafe { w.bits(0x3FF) });
+
+            needs_wake = true;
+        }
 
         self.state = match self.state {
             State::Idle => State::Idle,
             State::WaitForCommand => {
                 if smhc.smhc_mintsts.read().m_cc_int().bit_is_set() {
+                    smhc.smhc_rintsts.write(|w| w.cc().set_bit());
                     match self.op {
                         SmhcOp::None => State::Idle,
                         SmhcOp::Control => {
                             needs_wake = true;
                             State::Idle
                         }
-                        SmhcOp::Read { .. } | SmhcOp::Write { .. } => State::WaitForDataTransfer,
+                        SmhcOp::Read { .. } | SmhcOp::Write { .. } => State::WaitForDma,
                     }
+                } else {
+                    self.state
+                }
+            }
+            State::WaitForDma => {
+                // TODO: better way to check for RX and TX,
+                // *normal interrupt summary* does not seem to work
+                if smhc.smhc_idst.read().bits() != 0 {
+                    smhc.smhc_idst.write(|w| {
+                        w.nor_int_sum().set_bit();
+                        w.rx_int().set_bit();
+                        w.tx_int().set_bit();
+                        w
+                    });
+                    State::WaitForDataTransfer
                 } else {
                     self.state
                 }
             }
             State::WaitForDataTransfer => {
                 if smhc.smhc_mintsts.read().m_dtc_int().bit_is_set() {
-                    match self.op {
+                    smhc.smhc_rintsts.write(|w| w.dtc().set_bit());
+                    match &mut self.op {
                         SmhcOp::None | SmhcOp::Control => State::Idle,
-                        SmhcOp::Read { auto_stop, .. } | SmhcOp::Write { auto_stop, .. } => {
-                            if auto_stop {
+                        SmhcOp::Read {
+                            auto_stop,
+                            buf,
+                            cnt,
+                        }
+                        | SmhcOp::Write {
+                            auto_stop,
+                            buf,
+                            cnt,
+                        } => {
+                            tracing::trace!("setting buf len: {cnt}");
+                            // TODO: Safety?
+                            unsafe { buf.as_vec_mut().set_len(*cnt) };
+                            if *auto_stop {
                                 State::WaitForAutoStop
                             } else {
                                 needs_wake = true;
@@ -502,6 +648,7 @@ impl SmhcData {
             }
             State::WaitForAutoStop => {
                 if smhc.smhc_mintsts.read().m_acd_int().bit_is_set() {
+                    smhc.smhc_rintsts.write(|w| w.acd().set_bit());
                     needs_wake = true;
                     State::Idle
                 } else {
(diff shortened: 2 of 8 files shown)


source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,467 @@
...
+    /// Hardware specific configuration that should be applied
+    pub options: HardwareOptions,
+    /// The type of command
+    pub kind: CommandKind,
+    /// The expected type of the response
+    pub rsp_type: ResponseType,
+    /// Whether the response CRC needs to be checked
+    pub rsp_crc: bool,
+    /// Optional buffer for incoming or outgoing data
+    pub buffer: Option<FixedVec<u8>>,
```
jspngh, line 67: Here I'm not sure a `FixedVec` is the right thing to use.
In the blocking implementation of oreboot I'm using `&(mut) [Block]` (where a Block is 512 bytes),
but does not look feasible here, since you will probably run into issues with lifetime when passing this
from kernel to driver.

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,762 @@
...
+                            cnt,
+                        }
+                        | SmhcOp::Write {
+                            auto_stop,
+                            buf,
+                            cnt,
+                        } => {
+                            tracing::trace!("setting buf len: {cnt}");
+                            // TODO: Safety?
+                            unsafe { buf.as_vec_mut().set_len(*cnt) };
```
jspngh, line 636: One downside of using a `FixedVec`: having to manually set the length after you read the data.

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,762 @@
...
+                return Err(sdmmc::Error::Other);
+            }
+        };
+
+        // Statically allocate space for 16 DMA descriptors.
+        // Each descriptor can do a transfer of 4KB, giving a max total transfer of 64KB.
+        // By declaring them in this scope the descriptor memory will live long enough,
+        // however currently this is up to the user to guarantuee.
+        // Safety: a zero'ed descriptor is valid and will simply be ignored by the IDMAC.
+        let mut descriptors: [idmac::Descriptor; 16] = unsafe { core::mem::zeroed() };
```
jspngh, line 398: I'd like to improve this, but don't know if dynamically allocating memory for descriptors is the way to go.

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,762 @@
...
+                    descriptors[index] = idmac::Descriptor::try_from(idmac::DescriptorConfig {
+                        disable_int_on_complete: true,
+                        first: index == 0,
+                        last: false,
+                        buff_size: 0x1000,
+                        buff_addr: unsafe { buf_ptr.add(index * 0x1000) },
+                        next_desc: Some(&descriptors[index + 1] as *const _ as _),
+                    })
+                    .map_err(|_| sdmmc::Error::Other)?;
+                    remaining -= 0x1000;
```
hawkw, line 422: tiny style nit: maybe consider making 0x1000 a constant, like:
```suggestion
                const DESCR_BUFF_SIZE: u16 = 0x1000;
                while remaining > DESCR_BUFF_SIZE {
                    descriptors[index] = idmac::Descriptor::try_from(idmac::DescriptorConfig {
                        disable_int_on_complete: true,
                        first: index == 0,
                        last: false,
                        buff_size: DESCR_BUFF_SIZE,
                        buff_addr: unsafe { buf_ptr.add(index * DESCR_BUFF_SIZE) },
                        next_desc: Some(&descriptors[index + 1] as *const _ as _),
                    })
                    .map_err(|_| sdmmc::Error::Other)?;
                    remaining -= DESCR_BUFF_SIZE;
```

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,762 @@
...
+
+                descriptors[index] = idmac::Descriptor::try_from(idmac::DescriptorConfig {
+                    disable_int_on_complete: false,
+                    first: index == 0,
+                    last: true,
+                    buff_size: remaining as u16,
+                    buff_addr: unsafe { buf_ptr.add(index * 0x1000) },
+                    next_desc: None,
+                })
+                .map_err(|_| sdmmc::Error::Other)?;
```
hawkw, line 434: take it or leave it, up to you: we could get rid of the final assignment outside of the loop if we changed the loop a little bit:
```suggestion
                while remaining > 0 {
                    let buff_size = core::cmp::min(0x1000, remaining);
                    remaining = remaining.saturating_sub(0x1000);
                    let last = remaining == 0;
                    descriptors[index] = idmac::Descriptor::try_from(idmac::DescriptorConfig {
                        disable_int_on_complete: true,
                        first: index == 0,
                        last,
                        buff_size,
                        buff_addr: unsafe { buf_ptr.add(index * 0x1000) },
                        next_desc: (!last).then_some(|_| (&descriptors[index + 1] as *const _ as _)),
                    })
                    .map_err(|_| sdmmc::Error::Other)?;
                    index += 1;
                }
```
although i'm not sure how much clearer this is, really. up to you...

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,762 @@
...
+                Ok(sdmmc::Response::Short {
+                    value: self.smhc.smhc_resp0.read().bits(),
+                    data: match core::mem::replace(&mut guard.data.op, SmhcOp::None) {
+                        SmhcOp::Read { buf, .. } => Some(buf),
+                        _ => None,
+                    },
+                })
+            }
+        };
+        res
```
hawkw, line 487: style nit: i think this can be simplified a bit to just:
```suggestion
        if let Some(error) = guard.data.err.take() {
            tracing::warn!(?error, "SMHC error");
            Err(sdmmc::Error::Other) // TODO
        } else if long_resp {
            Ok(sdmmc::Response::Long([
                self.smhc.smhc_resp0.read().bits(),
                self.smhc.smhc_resp1.read().bits(),
                self.smhc.smhc_resp2.read().bits(),
                self.smhc.smhc_resp3.read().bits(),
            ]))
        } else {
            Ok(sdmmc::Response::Short {
                value: self.smhc.smhc_resp0.read().bits(),
                data: match core::mem::replace(&mut guard.data.op, SmhcOp::None) {
                    SmhcOp::Read { buf, .. } => Some(buf),
                    _ => None,
                },
            })
        }
```

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,762 @@
...
+
+            self.data.waker = Some(cx.waker().clone());
+            waiting = true;
+            self.smhc.smhc_ctrl.modify(|_, w| w.ine_enb().enable());
+
+            Poll::Pending
+        })
+        .await;
+    }
+}
```
hawkw, line 524: not something we need to address in this PR, but i kinda wonder if there's a nice way to factor out this pattern of an "IRQ lock", since it now shows up in two different driver impls...we could maybe add a trait for locking/unlocking an IRQ and implement the lock stuff on top of that?

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,762 @@
...
+            Some(ErrorKind::DataCrc)
+        } else if mint.m_dse_bc_int().bit_is_set() {
+            Some(ErrorKind::DataStart)
+        } else if mint.m_dee_int().bit_is_set() {
+            Some(ErrorKind::DataEnd)
+        } else if idst.des_unavl_int().bit_is_set() || idst.fatal_berr_int().bit_is_set() {
+            Some(ErrorKind::Dma)
+        } else {
+            None
+        }
```
hawkw, line 570: nit, take it or leave it: could potentially be a few lines shorter if this was, like, 
```rust
match () {
   _ if mint.whatever_bit().bit_is_set() => Some(ErrorKind::Whatever),
   // ...
}
```
although it's not a huge deal either way.

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,762 @@
...
+            .into_request_stream(queued)
+            .await;
+
+        kernel.spawn(self.run(rx)).await;
+        tracing::info!("SMHC driver task spawned");
+
+        Ok(())
+    }
+
+    #[tracing::instrument(name = "SMHC", level = tracing::Level::INFO, skip(self, rx))]
```
hawkw, line 329: tiny nit: can we include the SMHC peripheral number in this span, so we can tell whether it's SMHC0, SMHC1, or SMHC2?

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,762 @@
...
+/// Internal DMA controller
+mod idmac {
+    #[repr(C, align(4))]
+    #[derive(Default, Debug)]
+    pub struct Descriptor {
+        des0: u32,
+        des1: u32,
+        des2: u32,
+        des3: u32,
+    }
```
hawkw, line 679: take it or leave it: we _could_ use [my homemade bitfield library](https://docs.rs/mycelium-bitfield/latest/mycelium_bitfield/) for this, the way we do for the [normal DMAC descriptor bitfields](https://github.com/tosc-rs/mnemos/blob/b3b4da9d7c6aa4f1df104db9ea95e1c6aa757352/platforms/allwinner-d1/d1-core/src/dmac/descriptor.rs#L143-L196). 

the main advantage of this is we could have a nice builder-y interface where setting stuff actually just toggles the bit(s), rather than constructing a struct with a different layout and then using that to build the actual bitfield. but, since you already have a working implementation, don't feel like you need to rewrite it just to use my silly crate :)

platforms/allwinner-d1/src/lib.rs
```diff
@@ -155,6 +159,43 @@ pub fn kernel_entry(config: mnemos_config::MnemosConfig<PlatformConfig>) -> ! {
     #[cfg(feature = "sharp-display")]
     d1.initialize_sharp_display();
 
+    // #[cfg(feature = "sdcard")]
```
hawkw, line 162: should we either remove the feature flag, or put it back?

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,467 @@
...
+use crate::{
+    comms::{
+        kchannel::{KChannel, KConsumer, KProducer},
+        oneshot::{self, Reusable},
+    },
+    mnemos_alloc::containers::FixedVec,
+    registry::{self, known_uuids, Envelope, KernelHandle, RegisteredDriver},
+    Kernel,
+};
+use core::{convert::Infallible, fmt, time::Duration};
```
hawkw, line 18: clippy is mad about a bunch of unused imports in this file:
```suggestion
use crate::{
    comms::oneshot::Reusable,
    mnemos_alloc::containers::FixedVec,
    registry::{self, known_uuids, Envelope, KernelHandle, RegisteredDriver},
    Kernel,
};
```

source/kernel/src/services/sdmmc.rs
hawkw: i think fixedvec is _probably_ the correct thing to use for this, although i suppose it could also be a `FixedVec<Block>`?

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,467 @@
...
+/// Card identification register in R2 response format
+///   Manufacturer ID       [127:120]
+///   OEM/Application ID    [119:104]
+///   Product name          [103:64]
+///   Product revision      [63:56]
+///   Product serial number [55:24]
+///   Reserved              [23:20]
+///   Manufacturing date    [19:8]
+///   CRC7 checksum         [7:1]
+///   Not used, always 1    [0:0]
```
hawkw, line 188: rustdoc parses the bitfields in `[square brackets]` as links, and is complaining that they're broken. let's do something like:
```suggestion
///   Manufacturer ID       `[127:120]`
///   OEM/Application ID    `[119:104]`
///   Product name          `[103:64]`
///   Product revision      `[63:56]`
///   Product serial number `[55:24]`
///   Reserved              `[23:20]`
///   Manufacturing date    `[19:8]`
///   CRC7 checksum         `[7:1]`
///   Not used, always 1    `[0:0]`
```
so that it doesn't do that?

hawkw: also, we could maybe make this a Markdown table so it's formatted nicer in the rendered docs?
```suggestion
/// | Field                 | Bits        |
/// |:----------------------|------------:|
/// | Manufacturer ID       | `[127:120]` |
/// | OEM/Application ID    | `[119:104]` |
/// | Product name          | `[103:64]`  |
/// | Product revision      | `[63:56]`   |
/// | Product serial number | `[55:24]`   |
/// | Reserved              | `[23:20]`   |
/// | Manufacturing date    | `[19:8]`    |
/// | CRC7 checksum         | `[7:1]`     |
/// | Not used, always 1    | `[0:0]`     |
```

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,467 @@
...
+pub struct CardIdentification(u128);
```
hawkw, line 189: this could also potentially be a nice use for [`mycelium-bitfield`](https://docs.rs/mycelium-bitfield/latest/mycelium_bitfield/index.html) --- up to you though!

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,467 @@
...
+                rsp_type: ResponseType::Long,
+                rsp_crc: true,
+                buffer: None,
+            })
+            .await?
+        {
+            Response::Short { .. } => return Err(Error::Response),
+            Response::Long(value) => {
+                tracing::trace!("CMD2 response: {value:?}");
+                // TODO: map [u32; 4] value to u128
```
hawkw, line 342: let's make sure this happens before merging :)

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,467 @@
...
+                Ok(CardIdentification(0))
+            }
+        }
+    }
+
+    /// Get the relative card address
+    pub async fn get_rca(&mut self) -> Result<u32, Error> {
```
hawkw, line 349: i wonder if we should consider adding a `Rca(u32)` or `RelativeCardAddress(u32)` newtype, so that the user can only call methods like `select` and, `set_wide_bus`, `read`, etc, with real RCAs they obtained from this method, rather than random `u32`s that might not index a real card?

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,467 @@
...
+                tracing::trace!("CMD18 response: {value:#x}");
+                Ok(res)
+            }
+            _ => Err(Error::Response),
+        }
+    }
+}
+
+/// A client for SDIO cards using the [`SdmmcService`].
+pub struct SdioClient {
```
hawkw, line 458: let's maybe add 
```suggestion
#[allow(dead_code)]
pub struct SdioClient {
```
on this for now so that clippy stops whining about it?

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,467 @@
...
+        }
+    }
+
+    /// Read the desired number of data blocks into the provided buffer, starting at the given sector.
+    pub async fn read(
+        &mut self,
+        sector: u32,
+        blocks: usize,
+        buf: FixedVec<u8>,
+    ) -> Result<FixedVec<u8>, Error> {
```
hawkw, line 432: i think we might want to validate that  `buf.capacity()` is >= `blocks * 512` here and return an error if it isn't? the actual sdmmc service implementation does check that before configuring the DMA descriptors, but it seems like it could be worthwhile to check here as well, so that we can avoid sending the request if the driver is just going to immediately reject it for not having a long enough buffer...

also, it might be worth having an additional layer of defense against buffer overruns if another platform implementation's SDMMC service doesn't correctly check before reading...

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,467 @@
...
+            Response::Long(_) => Err(Error::Response),
+        }
+    }
+
+    /// Read the desired number of data blocks into the provided buffer, starting at the given sector.
+    pub async fn read(
+        &mut self,
+        sector: u32,
+        blocks: usize,
+        buf: FixedVec<u8>,
```
hawkw, line 431: if reads are always block-oriented, would it make sense to take a `FixedVec<[u8; 512]>` here, rather than a `FixedVec<u8>`? or do you think that just makes things more painful for the user of the API?

source/kernel/src/services/sdmmc.rs
hawkw, line 432: also, not something to address in this PR, but something that i was thinking about doing for all driver impls that consume a buffer to read/write into/from could maybe return it in their error type if the read/write failed (the i2c driver could also do this). this way, the user of the API can continue reusing their allocation, even if a particular read/write op failed. 

we could make this change later though.

hawkw (commented):
This is *extremely* cool, thanks for putting so much work into it! Overall, it looks great! 

I commented on some questions and minor suggestions I had while reading through the code. In general, though, this is great work and fits into the idiom for mnemOS drivers really nicely!

source/kernel/src/services/sdmmc.rs
jspngh: That's a nice idea, I'll experiment with this.

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
jspngh: I'm going  to look at the regular DMAC code and try to do things the same way for this internal DMAC.

platforms/allwinner-d1/d1-core/src/ccu.rs
```diff
@@ -20,10 +20,10 @@ pub struct Ccu {
 
 /// Trait to be implemented for module clocks that can be gated and reset
 pub trait BusGatingResetRegister {
-    /// Enable or disable the clock reset bit
-    fn gating(ccu: &mut CCU, pass: bool);
     /// Enable or disable the clock gating bit
-    fn reset(ccu: &mut CCU, deassert: bool);
+    fn gating(ccu: &mut CCU, pass: bool);
+    /// Enable or disable the clock reset bit
+    fn reset(ccu: &mut CCU, assert: bool);
```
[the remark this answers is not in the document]
jspngh: We didn't really get them backwards, but the initial API wasn't very intuitive.
When you passed `true` to reset, it *de-asserted* the reset (because that's what writing `1` in the register does).
I switched it around now, but using enums is probably the better idea.

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,762 @@
...
+        // Reset the FIFO
+        smhc.smhc_ctrl.modify(|_, w| w.fifo_rst().reset());
+        while smhc.smhc_ctrl.read().fifo_rst().is_reset() {
+            core::hint::spin_loop();
+        }
+
+        // Global interrupt disable
+        smhc.smhc_ctrl.modify(|_, w| w.ine_enb().disable());
+
+        prg_clk();
```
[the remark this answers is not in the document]
jspngh: I think I got this from the Linux of u-boot code.
I'll check if this is actually necessary :sweat_smile:

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,762 @@
...
+            .bind_konly::<SdmmcService>(queued)
+            .await?
+            .into_request_stream(queued)
+            .await;
+
+        kernel.spawn(self.run(rx)).await;
+        tracing::info!("SMHC driver task spawned");
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
jspngh: Difficult to answer right now. Once we get to implementing SDIO or MMC we should have a better idea if everything can be done in 1 `SdmmcService` or we need to make a split. At that point we can do some refactoring if necessary.

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
jspngh: It might make it a bit more complex, but I like it that everything is done inside the loop, so I'm going to apply this suggestion!

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
jspngh: Yeah, I think this pattern might show up a lot!

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
jspngh: That would be good. I guess the `Smhc` struct should then also include its *number* (0, 1, 2) to make this possible?

platforms/allwinner-d1/src/lib.rs
jspngh: I think adding a `sdcard` feature may come in handy. I still need to check how to do that though.
However, this function right here was just for testing, so I'm going to remove it before finalizing PR.

source/kernel/src/services/sdmmc.rs
jspngh: Good idea!

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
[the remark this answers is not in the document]
hawkw: well, i'm happy to keep this "just to be safe", too, i was just wondering if we understood why it was necessary. a comment saying "linux does this twice, so we do too, i guess" would be fine!

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
hawkw: yeah, adding a u8 for it seems fine to me!

jspngh:
I've been testing mostly with a lichee rv SOM, but have recently also started using a mango pi mq pro.
On the mango pi, I'm noticing some issues with initializing the SMHC, possibly the same as when I insert the lichee rv SOM into the dock board.
I'm going to try to fix this before implementing the rest of the feedback.

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
jspngh: Hey @hawkw, could you point me in the right direction regarding the way descriptors are allocated for the regular DMAC? I don't find references to `mnemos_bitpool` in the code and from what I can see the two current DMAC users (uart + spim) have enough with 1 descriptor per transfer (whereas here the number of required descriptors depends on the size of the transfer).

hawkw:
> On the mango pi, I'm noticing some issues with initializing the SMHC, possibly the same as when I insert the lichee rv SOM into the dock board.
> I'm going to try to fix this before implementing the rest of the feedback.

Sounds good, thanks for testing with both boards! <3

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,754 @@
...
+                    remaining = remaining.saturating_sub(DESCR_BUFF_SIZE);
+                    let first = index == 0;
+                    let last = remaining == 0;
+                    descriptors[index] = idmac::Descriptor::try_from(idmac::DescriptorConfig {
+                        disable_int_on_complete: !last,
+                        first,
+                        last,
+                        buff_size,
+                        buff_addr: unsafe { buf_ptr.add(index * DESCR_BUFF_SIZE) },
+                        next_desc: (!last).then_some(&descriptors[index + 1] as *const _ as _),
```
hawkw, line 425: hmm, can this bounds check fail if we're on the last descriptor? should we maybe be using 
```suggestion
                        next_desc: (!last).then(|| &descriptors[index + 1] as *const _ as _),
```
so that the indexing only occurs if `last` is `false`?

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
hawkw: The code I was referring to is the DMAC channel allocator in [`d1_core::dmac`](https://github.com/tosc-rs/mnemos/blob/main/platforms/allwinner-d1/d1-core/src/dmac/mod.rs), particularly the [`Dmac::claim_channel`](https://mnemos.dev/doc/mnemos_d1_core/dmac/struct.dmac#method.claim_channel) method and friends. I think a somewhat similar pattern could be used here, potentially.

I probably wouldn't worry about that right now though. If the current solution works, this might be worth saving for a subsequent PR.

platforms/allwinner-d1/d1-core/src/ccu.rs
```diff
@@ -248,12 +260,24 @@ macro_rules! impl_bgr {
...
 
-                fn reset(ccu: &mut CCU, deassert: bool) {
-                    ccu.$reg.modify(|_, w| w.$reset().bit(deassert));
+                fn reset(ccu: &mut CCU, reset: BusReset) {
+                    ccu.$reg.modify(|_, w| {
+                        if reset == BusReset::Assert {
+                            w.$reset().clear_bit()
+                        } else {
+                            w.$reset().set_bit()
+                        }
```
hawkw, line 279: nit, take it or leave it: this could still be a one-liner if we wrote
```suggestion
                        w.$reset.bit(reset == BusReset::Deassert);
```

platforms/allwinner-d1/d1-core/src/ccu.rs
```diff
@@ -248,12 +260,24 @@ macro_rules! impl_bgr {
...
             impl BusGatingResetRegister for $MODULE {
-                fn gating(ccu: &mut CCU, pass: bool) {
-                    ccu.$reg.modify(|_, w| w.$gating().bit(pass));
+                fn gating(ccu: &mut CCU, gating: BusGating) {
+                    ccu.$reg.modify(|_, w| {
+                        if gating == BusGating::Mask {
+                            w.$gating().clear_bit()
+                        } else {
+                            w.$gating().set_bit()
+                        }
```
hawkw, line 269: nit, take it or leave it: this could still be a one-liner if we wrote
```suggestion
                        w.$gating.bit(gating == BusGating::Pass);
```

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,754 @@
...
+        // Set number of bytes that will be read or written.
+        self.smhc.smhc_bytcnt.write(|w| unsafe { w.bits(byte_cnt) });
+
+        // Soft reset of DMA controller
+        self.smhc.smhc_idmac.write(|w| w.idmac_rst().set_bit());
+
+        // Configure the burst size and TX/RX trigger level
+        self.smhc.smhc_fifoth.write(|w| {
+            w.tx_tl().variant(8);
+            w.rx_tl().variant(7);
```
hawkw, line 275: minor nit: can we add what burst size and trigger levels were selecting to the comment? it's not immediately clear what "8" and "7" are saying here — is that the actual value of the trigger level or is it selecting something else?

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,754 @@
...
+            w.idmac_enb().set_bit();
+            w.fix_bust_ctrl().set_bit();
+            w
+        });
+
+        // Reset the FIFO
+        self.smhc.smhc_ctrl.modify(|_, w| w.fifo_rst().reset());
+        while self.smhc.smhc_ctrl.read().fifo_rst().is_reset() {
+            core::hint::spin_loop();
+        }
```
hawkw, line 299: very minor nit, not actually important: we have very similar code in `Smhc::init`, would it make sense to factor this out into a 
```rust
fn reset_fifo(regs: &smhc::RegisterBlock) {
    // ...
}
```
or something?

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,754 @@
...
+    /// The descriptor chain needs to live at least as long as the DMA transfer.
+    /// Additionally, their content (e.g., the validity of the buffers they point to)
+    /// also needs to be verified by the user.
+    unsafe fn prepare_dma(&self, descriptor: &idmac::Descriptor, byte_cnt: u32) {
+        self.smhc.smhc_ctrl.modify(|_, w| {
+            w.dma_enb().set_bit();
+            w.dma_rst().set_bit();
+            w
+        });
+        while self.smhc.smhc_ctrl.read().dma_rst().bit_is_set() {}
```
hawkw, line 258: should this loop have a spin loop hint in it, like we do elsewhere?
```suggestion
        while self.smhc.smhc_ctrl.read().dma_rst().bit_is_set() {
             core::hint::spin_loop();
         }
```

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,754 @@
...
+        match &guard.data.op {
+            SmhcOp::Read { buf, cnt, .. } | SmhcOp::Write { buf, cnt, .. } => {
+                const DESCR_BUFF_SIZE: usize = 0x1000;
+
+                // Currently we limit the number of data that can be read at once
+                if *cnt > DESCR_BUFF_SIZE * descriptors.len() || buf.capacity() < *cnt {
+                    return Err(sdmmc::Error::Data);
+                }
+
+                tracing::debug!("Creating descriptor chain from buffer");
```
hawkw, line 410: nit, take it or leave it: maybe worth including the descriptor count?
```suggestion
                tracing::debug!(cnt, "Creating descriptor chain from buffer");
```

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,754 @@
...
+                        }
+                        SmhcOp::Read { .. } | SmhcOp::Write { .. } => State::WaitForDma,
+                    }
+                } else {
+                    self.state
+                }
+            }
+            State::WaitForDma => {
+                // TODO: better way to check for RX and TX,
+                // *normal interrupt summary* does not seem to work
```
hawkw, line 598: out of curiousity, what does this mean exactly?

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,754 @@
...
+                        w
+                    });
+                    State::WaitForDataTransfer
+                } else {
+                    self.state
+                }
+            }
+            State::WaitForDataTransfer => {
+                if smhc.smhc_mintsts.read().m_dtc_int().bit_is_set() {
+                    smhc.smhc_rintsts.write(|w| w.dtc().set_bit());
```
hawkw, line 613: should this be a `modify` instead of a `write`? the write will set the register to _only_ have the DTC bit set, is that what we want? or do we only want to touch that bit and leave the others unchanged?

if we do want to clear every other bit, a comment stating that could be nice...

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,754 @@
...
+                            }
+                        }
+                    }
+                } else {
+                    self.state
+                }
+            }
+            State::WaitForAutoStop => {
+                if smhc.smhc_mintsts.read().m_acd_int().bit_is_set() {
+                    smhc.smhc_rintsts.write(|w| w.acd().set_bit());
```
hawkw, line 643: similarly, this clears every bit besides ACD, is that desirable?

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,754 @@
...
+            }
+        }
+    }
+}
+
+/// Internal DMA controller
+mod idmac {
+    #[repr(C, align(4))]
+    #[derive(Default, Debug)]
+    pub struct Descriptor {
```
hawkw, line 666: nit, TIOLI: this struct (and everything else in the `idmac` module) really only needs to be 
```suggestion
    pub(super) struct Descriptor {
```
since it's never re-exported...

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
jspngh: In the IDMAC interrupt status register has the `normal interrupt summary` as bit 8, which is supposed to be set when either of the RX or TX interrupt bits are set. That would be convenient to check if the operation we are waiting for has finished.
However, it doesn't work and the bit is never set.
Maybe the documentation of the interrupt enable register isn't complete in the manual version I have (v1.0).
If you have the newer version, could you let me know if `SMHC_IDIE_REG` has a bit 8 that can be set? :smile:

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
jspngh: This is indeed a bit counterintuitive, but many of the interrupt registers in the D1 are "write 1 to clear".
Which means that this write sets *only* the DTC bit back to 0 and the rest are left untouched.
Using `modify` here would actually reset other bits that are also set, which is undesirable.
I'll add a general comment about it in the `advance_isr` function.

hawkw: cool, thank you! let's definitely add the comment.

in other drivers, we do generally use `modify`, as it will write back every bit that's currently 1 as a 1, which _should_ be sufficient to clear the interrupt I think --- cc @jamesmunns

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
hawkw: i checked v1.4 of the D1 manual, and the table for `SMHC_IDIE_REG` lists the following bits:

| Bit | Name |
|:-----|:---------|
| 0 | `TX_INT_ENB` |
| 1 | `RX_INT_ENB` |
| 2 | `FERR_INT_ENB` |
| 4 | `DES_UNAVL_INT_ENB` |
| 5| `ERR_SUM_INT_ENB` |

Bit 3 and bits 6-31 are not listed in the table in that version of the manual. I assume this means that the "normal interrupt summary" bit is not actually real?

hawkw: Ah, it looks like v1.4 of the manual _does_ discuss a `NOR_INT_SUM` ("Normal Interrupt Summary") bit, but it's bit 8 of the `SMHC_IDST_REG` register, rather than `SMHC_IDIE_REG`.

The documentation for that bit says the following:
> NOR_INT_SUM (NIS)
>
> Normal Interrupt Summary
>
> Logical OR of the following:
> IDST[0]: Transmit Interrupt
> IDST[1]: Receive Interrupt
>
> Only unmasked bits affect this bit.
> This is a sticky bit and must be cleared each time a corresponding
> bit that causes NIS to be set is cleared.

Hope that's helpful!

jspngh: Ah okay, it's the same for the v1.0 manual. The IDST is the status register while the IDIE is the enable register.
Since the IDST talks about bit 8, but I don't see it getting set, I was wondering if it had to be enabled as well, and that the corresponding bit in the IDIE was missing in my manual version. But it doesn't seem so.

hawkw: Oh, I see. Hmm. The Allwinner docs are....not great... I wonder if they're trying to say that you need to be writing it back every time we read from the IDST reg to ensure it gets set on the next IRQ?

jspngh:
> > On the mango pi, I'm noticing some issues with initializing the SMHC, possibly the same as when I insert the lichee rv SOM into the dock board.
> > I'm going to try to fix this before implementing the rest of the feedback.
> 
> Sounds good, thanks for testing with both boards! <3

I think my mango pi might already be busted :frowning_face: 
`xfel ddr d1` gives this output:

```sh
DRAM only have internal ZQ!!
get_pmu_exist() = 4294967295
ddr_efuse_type: 0x0
[AUTO DEBUG] two rank and full DQ!
ddr_efuse_type: 0x0
[AUTO DEBUG] rank 0 row = 16 
[AUTO DEBUG] rank 0 bank = 8 
[AUTO DEBUG] rank 0 page size = 8 KB 
[AUTO DEBUG] rank 1 row = 15 
[AUTO DEBUG] rank 1 bank = 8 
[AUTO DEBUG] rank 1 page size = 2 KB 
rank1 config different from rank0
DRAM BOOT DRIVE INFO: %s
DRAM CLK = 792 MHz
DRAM Type = 3 (2:DDR2,3:DDR3)
DRAMC ZQ value: 0x7b7bfb
DRAM ODT value: 0x42.
ddr_efuse_type: 0x0
DRAM SIZE =4608 M
DRAM simple test FAIL.
fedcc098 != fedcba98 at address d0000000
```

hawkw:
> > > On the mango pi, I'm noticing some issues with initializing the SMHC, possibly the same as when I insert the lichee rv SOM into the dock board.
> 
> > > I'm going to try to fix this before implementing the rest of the feedback.
> 
> > 
> 
> > Sounds good, thanks for testing with both boards! <3
> 
> 
> 
> I think my mango pi might already be busted :frowning_face: 
> 
> `xfel ddr d1` gives this output:
> 
> 
> 
> ```sh
> 
> DRAM only have internal ZQ!!
> 
> get_pmu_exist() = 4294967295
> 
> ddr_efuse_type: 0x0
> 
> [AUTO DEBUG] two rank and full DQ!
> 
> ddr_efuse_type: 0x0
> 
> [AUTO DEBUG] rank 0 row = 16 
> 
> [AUTO DEBUG] rank 0 bank = 8 
> 
> [AUTO DEBUG] rank 0 page size = 8 KB 
> 
> [AUTO DEBUG] rank 1 row = 15 
> 
> [AUTO DEBUG] rank 1 bank = 8 
> 
> [AUTO DEBUG] rank 1 page size = 2 KB 
> 
> rank1 config different from rank0
> 
> DRAM BOOT DRIVE INFO: %s
> 
> DRAM CLK = 792 MHz
> 
> DRAM Type = 3 (2:DDR2,3:DDR3)
> 
> DRAMC ZQ value: 0x7b7bfb
> 
> DRAM ODT value: 0x42.
> 
> ddr_efuse_type: 0x0
> 
> DRAM SIZE =4608 M
> 
> DRAM simple test FAIL.
> 
> fedcc098 != fedcba98 at address d0000000
> 
> ```

welp. that seems bad...

jspngh:
> welp. that seems bad...

No luck trying to revive it.

But in the meantime I've refactored the idmac code to also use a builder and bitfields.
Can you have a  look @hawkw to see if it's what you expected?

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,874 @@
...
+            /// Bits 29:5 are reserved.
+            const _RESERVED_1 = 25;
+
+            /// When some errors happen in transfer, this bit will be set to 1 by the IDMAC.
+            const ERR_FLAG = 1;
+
+            /// When set to 1, this bit indicates that the descriptor is owned by the IDMAC.
+            /// When this bit is reset, it indicates that the descriptor is owned by the host.
+            /// This bit is cleared when the transfer is over.
+            const DES_OWN_FLAG = 1;
```
hawkw, line 766: nit, take it or leave it: i would probably use `bool` for the various 1-bit flags here, rather than a single-bit numeric bitfield:
```suggestion
            const _RESERVED_0 = 1;

            /// Disable interrupts on completion.
            const CUR_TXRX_OVER_INT_DIS = 1;

            /// When set to 1, this bit indicates that the buffers this descriptor points to
            /// are the last data buffer.
            const LAST_FLAG: bool;

            /// When set to 1, this bit indicates that this descriptor contains
            /// the first buffer of data. It must be set to 1 in the first DES.
            const FIRST_FLAG: bool;

            /// When set to 1, this bit indicates that the second address in the descriptor
            /// is the next descriptor address. It must be set to 1.
            const CHAIN_MOD bool;

            /// Bits 29:5 are reserved.
            const _RESERVED_1 = 25;

            /// When some errors happen in transfer, this bit will be set to 1 by the IDMAC.
            const ERR_FLAG: bool;

            /// When set to 1, this bit indicates that the descriptor is owned by the IDMAC.
            /// When this bit is reset, it indicates that the descriptor is owned by the host.
            /// This bit is cleared when the transfer is over.
            const DES_OWN_FLAG:  bool;
```

stylistically, i would also probably drop `_FLAG` from these names, but that's up to you.

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,874 @@
...
+                ..self
+            }
+        }
+
+        pub fn last(self, val: bool) -> Self {
+            Self {
+                cfg: self.cfg.with(Cfg::LAST_FLAG, val as u32),
+                ..self
+            }
+        }
```
hawkw, line 810: since this builder type is only used in one place, i'm not convinced that we need methods for these...we could just have the code that builds the descriptor call `cfg.with(Cfg::CUR_TXRX_OVER_INT_DIS, ...).with(Cfg::FIRST_FLAG, ...)...`

hawkw (commented):
the added `mycelium-bitfield` types look nice! i had some minor suggestions.

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
jspngh: Do you mean adding them as parameters to `DescriptorBuilder::build`?
```rs
pub fn build(self, disable_irq: bool, first: bool, last: bool) -> Descriptor {
    Descriptor {
        configuration: self.cfg.with(...).with(...), // set all the required flags
        ...
    }
}
```
Only taking `self` in `build` feels nicer though imo.

jspngh: I'll keep it like it is for now then, we could always later change it to expose the `cfg` field.

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,489 @@
...
+                rsp_type: ResponseType::Long,
+                rsp_crc: true,
+                buffer: None,
+            })
+            .await?
+        {
+            Response::Short { .. } => Err(Error::Response),
+            Response::Long(value) => {
+                tracing::trace!("CMD2 response: {value:?}");
+                let mut bytes = [0u8; 16];
```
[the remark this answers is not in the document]
hawkw: i mean, they're laid out the same in memory so ideally the conversion should be a no-op...we could do a bunch of shifts and OR them all together, but that almost certainly won't compile down to a no-op. 

the most efficient solution would be to use unsafe code (either `mem::transmute` or a `union`) to guarantee that the conversion is a no-op, but we need to write the `unsafe` word to do that. if we were going to go down the unsafe route, i would prefer a union personally, since it doesn't allow totally unrestricted type punning. something like:
```rust
    Response::Long(value) => {
         tracing::trace!("CMD2 response: {value:?}");
         union CidResponse {
              value: [u16; 4]
              cid: u128
         }

        let rsp = CidResponse { value };
         unsafe {
             rsp.cid
         }
    }
```

although, on the other hand, using `mem::transmute` is significantly shorter.

it's up to you whether you think it's worth writing unsafe code for this. i don't think there's any danger in using `unsafe` here, but it's not strictly necessary. it does seem unfortunate to do a bunch of memcpys when the `[u32; 4]` and `u128` should be exactly the same in memory, though...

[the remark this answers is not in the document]
jspngh: I'll just go ahead and use your union example, since there's no clear "best" way of doing it.

hawkw:
hmm, CI failure looks kinda weird, i've never seen it do that before: https://github.com/tosc-rs/mnemos/actions/runs/7008859631/job/19066205659?pr=299#step:7:876

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
```diff
@@ -0,0 +1,890 @@
...
+                ErrorKind::DataCrc => sdmmc::Error::Data,
+                ErrorKind::ResponseTimeout => sdmmc::Error::Timeout,
+                ErrorKind::DataTimeout => sdmmc::Error::Timeout,
+                ErrorKind::DataStarvationTimeout => sdmmc::Error::Timeout,
+                ErrorKind::FifoUnderrunOverflow => sdmmc::Error::Data,
+                ErrorKind::CommandBusyIllegalWrite => sdmmc::Error::Busy,
+                ErrorKind::DataStart => sdmmc::Error::Data,
+                ErrorKind::DataEnd => sdmmc::Error::Data,
+                ErrorKind::Dma => sdmmc::Error::Other,
+                ErrorKind::Other => sdmmc::Error::Other,
```
hawkw, line 511: hmm, it's a bit unfortunate that some potentially useful information is lost here when converting these `ErrorKind`s into `sdmmc::Error`s. for example, the user cannot determine whether a `sdmmc::Error::Timeout` represents a `ResponseTimeout`, a `DataTimeout`, or a `DataStarvationTimeout`.

what do you think about changing the `sdmmc::Error` type to something like this:
```rust
pub struct Error {
    kind: sdmmc::ErrorKind,
    message: Option<&'static str>, 
}
```
(where `sdmmc::ErrorKind` is the enum currently named `sdmmc::Error`).

this way, users can still programmatically match on the enum, representing a standardized enumeration of possible kinds of errors that a SDMMC driver may encounter, but there's also the ability for the driver implementation to include extra human-readable information that can be printed out by the user, such as "response timeout" vs "data timeout", or disambiguating "data start" and "data end", etc?

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,489 @@
...
+                rsp_type: ResponseType::Short,
+                rsp_crc: true,
+                buffer: None,
+            })
+            .await?;
+
+        match self
+            .cmd(Command {
+                index: 6,
+                argument: 0b10,
```
hawkw, line 424: what does 0b10 mean here?

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,489 @@
...
+                options: HardwareOptions::None,
+                kind: CommandKind::Control,
+                rsp_type: ResponseType::Short,
+                rsp_crc: true,
+                buffer: None,
+            })
+            .await?
+        {
+            Response::Short { value, .. } => {
+                tracing::trace!("CMD8 response: {value:#x}");
```
hawkw, line 276: take it or leave it, not a big deal: a similar trace is included in each of the methods that send a command to the card...what do you think about adding this to the `cmd` method, instead, like:
```rust
     tracing::trace!("CMD{index}: response {value:#}");
```
so that we don't have to repeat it in every other method?

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,489 @@
...
+
+    /// Reset the card
+    pub async fn reset(&mut self) -> Result<(), Error> {
+        self.cmd(Command::default()).await.map(|_| ())
+    }
+
+    /// Initialize the card
+    pub async fn initialize(&mut self) -> Result<CardType, Error> {
+        /// Request switch to 1.8V
+        #[allow(dead_code)]
```
hawkw, line 252: where's the dead code here?

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,489 @@
...
+                    tracing::trace!("ACMD41 response: {value:#x}");
+                    if value & OCR_NBUSY == OCR_NBUSY {
+                        // Card has finished power up, data is valid
+                        break value;
+                    }
+                }
+                Response::Long(_) => return Err(Error::Response),
+            }
+
+            // TODO: wait some time (e.g., 1ms?)
```
hawkw, line 325: we do have a globally available kernel timer...

hawkw (approved):
this PR looks great and i'm happy to merge it whenever! i had some minor suggestions, but it's fine to address them later, or ignore me entirely :)

hawkw:
PR  #300 fixes the CI issues, so updating from `main` should get us a green build again.

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,489 @@
...
+        {
+            Response::Short { .. } => Err(Error::Response),
+            Response::Long(value) => {
+                tracing::trace!("CMD2 response: {value:?}");
+                union CidResponse {
+                    value: [u32; 4],
+                    cid: u128,
+                }
+                let rsp = CidResponse { value };
+                Ok(CardIdentification(unsafe { rsp.cid }))
```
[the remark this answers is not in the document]
jspngh: Looking at this again, it indeed makes sense to let the hardware specific driver do the conversion.
This is currently `[u32; 4]` because in the D1 you need to retrieve 4 32-bit registers to reconstruct the "long" response, but this shouldn't be part of the service interface.

source/kernel/src/services/sdmmc.rs
jspngh: For this command, you can use either `0b00` as argument, to tell the card to switch to 1-bit width or `0b10` to switch to 4-bit width. I added a comment to clarify.
I guess this `set_wide_bus` function should have been `set_bus_width` though, with an argument that specifies the width.
If it's okay for you, I'd prefer to postpone this change to a future PR though.

source/kernel/src/services/sdmmc.rs
jspngh: That would indeed be better. However, I guess you would have to match on the response in `cmd` then?
Or derive `Debug` on `Response`, which is probably the easiest solution.

source/kernel/src/services/sdmmc.rs
jspngh: The `OCR_S18R` constant is currently not used, but I wanted to keep the option open to start using it in the future without having to look up the value again / forgetting it exists :sweat_smile:.

source/kernel/src/services/sdmmc.rs
jspngh: Is this using `maitake::time::sleep(...).await`?

platforms/allwinner-d1/d1-core/src/drivers/smhc.rs
jspngh: Yeah, I like the idea. It's unfortunate that this
```rust
return Err(ErrorKind::Response)
```
needs to become something like this then
```rust
return Err(Error { kind: ErrorKind::Response, message: None })
```
or
```rust
return Err(Error::new(ErrorKind::Response))
```
in that case, although I guess the last option isn't too bad.

source/kernel/src/services/sdmmc.rs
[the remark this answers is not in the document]
hawkw: yeah, i think that's the right thing --- from the perspective of the hardware-agnostic `SdmmcService`, it's just a 128-bit number, i think. how that's constructed from register values or whatever is up to the driver implementation.

source/kernel/src/services/sdmmc.rs
hawkw: :woman_shrugging: if it's too much extra work, i'm fine leaving this as is!

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,523 @@
...
+impl core::fmt::Display for Error {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "SD/MMC error: ")?;
+        match self.kind {
+            ErrorKind::Busy => write!(f, "busy "),
+            ErrorKind::Response => write!(f, "response "),
+            ErrorKind::Data => write!(f, "data "),
+            ErrorKind::Buffer => write!(f, "buffer "),
+            ErrorKind::Timeout => write!(f, "timeout "),
+            ErrorKind::Other => write!(f, "other "),
```
hawkw, line 193: tiny style nit, not a blocker: i try to avoid using the `write!` macro when no formatted arguments are interpolated into the string. it's up to you whether you actually care about this, though... 
```suggestion
            ErrorKind::Data => f.write_str("data "),
            ErrorKind::Buffer => f.write_str("buffer "),
            ErrorKind::Timeout => f.write_str("timeout "),
            ErrorKind::Other => f.write_str("other "),
```

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,523 @@
...
+
+        self.message.map_or_else(|| Ok(()), |msg| msg.fmt(f))
+    }
+}
+
+impl core::fmt::Debug for Response {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Response::Short { value, data: _ } => write!(f, "{:#x}", value),
+            Response::Long(value) => write!(f, "{:#x}", value),
```
hawkw, line 204: tiny style nit, feel free to ignore me:
```suggestion
            Response::Short { value, data: _ } => write!(f, "{value:#x}"),
            Response::Long(value) => write!(f, "{value:#x}"),
```

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,523 @@
...
+    }
+
+    async fn cmd(&mut self, cmd: Command) -> Result<Response, Error> {
+        let index = cmd.index;
+        let result = self
+            .handle
+            .request_oneshot(cmd, &self.reply)
+            .await
+            .map_err(|error| {
+                tracing::warn!(?error, "failed to send request to SD/MMC service");
```
hawkw, line 293: nit: it could maybe be nice to include the command here as well:
```suggestion
                tracing::warn!(?error, "failed to send CMD{index} request to SD/MMC service");
```

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,523 @@
...
+                Error::from(ErrorKind::Other) // TODO
+            })
+            .and_then(|resp| resp.body);
+        tracing::trace!("CMD{} response: {:?}", index, result);
```
hawkw, line 297: extremely unimportant style nit, feel free to ignore this:
```suggestion
        tracing::trace!("CMD{index} response: {result:?}");
```

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,523 @@
...
+                options: HardwareOptions::None,
+                kind: CommandKind::Control,
+                rsp_type: ResponseType::Short,
+                rsp_crc: true,
+                buffer: None,
+            })
+            .await?
+        {
+            Response::Short { value, .. } => {
+                if value == 0x1AA {
```
hawkw, line 333: style nit, not a blocker: should `0x1AA` maybe be a const, since it's also the command index?

source/kernel/src/services/sdmmc.rs
```diff
@@ -0,0 +1,523 @@
...
+        if buf.capacity() < 512 * blocks {
+            return Err(Error::from(ErrorKind::Buffer));
+        }
+
+        match self
+            .cmd(Command {
+                index: 18,
+                argument: sector,
+                options: HardwareOptions::None,
+                kind: CommandKind::Read(512 * blocks),
```
hawkw, line 496: nit, not a blocker: maybe:
```suggestion
        const BLOCK_SIZE: usize = 512;
        
        let bytes = BLOCK_SIZE * blocks;
        // The provider buffer should have space for the requested amount of data
        if buf.capacity() < bytes {
            return Err(Error::from(ErrorKind::Buffer));
        }

        match self
            .cmd(Command {
                index: 18,
                argument: sector,
                options: HardwareOptions::None,
                kind: CommandKind::Read(bytes),
```

hawkw (approved):
this looks wonderful! i commented on some last very minor style suggestions i noticed while skimming your latest changes, but none of them are actually important. please go ahead and merge whenever you feel like this is ready, i'm really excited to get it merged! :D

PR #312, opened 2024-04-09
PLIC refactors and performance improvements

hawkw:
Currently, the interrupt vector dispatch code uses a linear search over
the interrupt vector array by checking each vector's ID against the
interrupt number being dispatched. This is inefficient; because we
already generate the vector array in order, we can simply index into it,
which is _O_(1) instead of _O_(_n<sub>interrupts</sub>_). This commit
changes the existing code to do that.

We can rely on the ordering being correct as it's generated by a `const
fn` that should always output the same order. However, I've also added
debug assertions that the index and ID match, just in case the table
somehow gets corrupted or something. I don't think this is really that
necessary, but it seemed nice to have.

### DIFF
--- a/platforms/allwinner-d1/d1-core/src/plic.rs
+++ b/platforms/allwinner-d1/d1-core/src/plic.rs
@@ -54,16 +54,14 @@ impl Plic {
     ///
     /// May effect normal interrupt processing
     pub unsafe fn unmask(&self, interrupt: Interrupt) {
-        let nr = interrupt.into_bits() as usize;
-        let (reg_offset, irq_en) = (nr / 32, 1 << (nr % 32));
-        self.plic.mie[reg_offset].modify(|r, w| w.bits(r.bits() | irq_en));
+        let (mie, irq_en) = self.index_mie(interrupt);
+        mie.modify(|r, w| w.bits(r.bits() | irq_en));
     }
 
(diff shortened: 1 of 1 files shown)


hawkw:
@jamesmunns I'd love to get your review on this, since I'm hacking up your code. Also, I haven't actually tested this on my D1 yet, so I should make sure it like, works.

hawkw:
hmm, i flashed this change on my D1 and observed the following:

- the UART doesn't seem to work (no output in crowtty)
- the blue blinky light is solid on rather than blinky

which kind of suggests that interrupt handling is broken. 

however, the keyboard and SHARP display _do_ seem to work. going to try flashing with a build from master and making sure we haven't just broken stuff on master too.

hawkw:
> hmm, i flashed this change on my D1 and observed the following:
> 
>     * the UART doesn't seem to work (no output in crowtty)
> 
>     * the blue blinky light is solid on rather than blinky
> 
> 
> which kind of suggests that interrupt handling is broken.
> 
> however, the keyboard and SHARP display _do_ seem to work. going to try flashing with a build from master and making sure we haven't just broken stuff on master too.

**Update**: nevermind, the "broken UART" was just "i was looking at the wrong USB TTY", and after rebasing onto `main I now see a proper blinkenlight. False alarm!

PR #325, opened 2024-08-10
CI script: use jq to filter and output final format

spookyvision:
this PR uses jq's (admittedly quite weird) builtin capabilities to create a list of crates to build docs for, instead of the previous grep/cut/sed/tr pipe.

### DIFF
--- a/scripts/run-ci-build.sh
+++ b/scripts/run-ci-build.sh
@@ -1,20 +1,12 @@
 #!/bin/bash
 
+# TODO -euo pipefail isn't optimal but I haven't got the energy to improve further -Anatol
 set -euxo pipefail
 
+# the old filter removed crowtty and manganese, but manganese isn't in deps anymore
 defaultmembers=$( \
     cargo metadata --format-version 1 | \
-    jq .workspace_default_members | \
-    grep -E '  ".*' | \
-    # crowtty's dependencies can't easily be installed on netlify
-    grep -v 'crowtty' | \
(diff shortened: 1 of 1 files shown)


scripts/run-ci-build.sh
```diff
@@ -1,20 +1,14 @@
...
-    # crowtty's dependencies can't easily be installed on netlify
-    grep -v 'crowtty' | \
-    # manganese depends on `libudev` (transitive dep via one of its' bindeps)
-    # which can't be installed on CI.
-    grep -v 'manganese' | \
-    cut -d" " -f3 | \
-    cut -d'"' -f2 | \
-    sed -E 's/(.*)/-p \1 /g' | \
-    tr -d '\n' \
+    jq -r '.workspace_default_members | del(.[] | select(contains("crowtty"))) | to_entries[] |"-p \(.value)"'
```
hawkw, line 11: hmm, this can't easily be made multi-line, i assume?

hawkw (approved):
i don't have a super strong grasp of `jq`, but this does seem less flaky than the previous sed/cut nonsense i wrote! thanks!

scripts/run-ci-build.sh
spookyvision: pushed a fix that seems to work

spookyvision:
@jamesmunns (low importance) if you feel like giving this a quick glance

hawkw:
I'm fine with just merging this whenever, but if you want to wait for @jamesmunns, that's fine too. I'm pretty sure the new code does the right thing...

PR #334, opened 2024-09-29
tools: rewrite x86_64 build system

### LINKED ISSUE #332: x86 bootimage runner is bad and we should feel bad too
so it turns out that the bootimage stuff for x86 is totally busted in a REALLY INFURIATING WAY: if you have built a kernel image and it's in your target dir, it will just run that. if you change the kernel and you run the x86 bootimage runner again, it will happily launch QEMU for you and RUN THE SAME KERNEL AGAIN instead of building you a new one. this sucks and is dumb and i hate myself for doing it.

**TIME TO REWRITE THE BUILD SYSTEM AGAIN I GUESS LOL**

hawkw:
So I rewrote the x86_64 kernel's build system again.

This commit rewrites the builder for the x86_64 kernel to be a cargo
custom runner, rather than a build script in a normal crate. This way,
we can use

```console
$ cargo run --package mnemos-x86_64 --target-arch x86_64-unknown-none
```

to build the x86 kernel normally using cargo, outputting a path that's
then passed into the custom runner binary by Cargo. This way, we don't
build the whole kernel inside a build script, so we get normal cargo
output from the kernel build, and we don't need to deal with the fact
that we have to rerun the build script whenever the kernel or any of its
dependencies changes. Instead, the kernel is built by a normal cargo
invocation, which is like infinitely better in literally every possible
way. The custom runner can also launch QEMU, so you can now `cargo run`
the x86 kernel and IT JUST WORKS, RIGHT OUT OF THE BOX.[^1]

The one weird thing about this approach is that now, you build the
kernel by running `cargo run` with a `build` subcommand that tells it to
not actually build the kernel, instead of `cargo build` like a normal
person. This is, admittedly, weird; but it's so much less bad than the
previous thing where you had to run `cargo clean` every time you changed
the kernel if you didn't want to get a stale build artifact (see #332),
which was *incredibly life-ruining* and made developing the x86 kernel a
horrific nightmare of infinite pain. And, most of the time you'll build
it using the Justfile `just run-x86` and `just build-x86` recipes so you
don't have to face the horrible truth that you're using `cargo run` to
build something without running it. So, whatever.

Closes #332

[^1]: Provided that, like, you have `qemu-system-x86_64` on your
    computer, and stuff.


tools/x86_64-bootimager/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+impl fmt::Display for BootLogLevel {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(match self {
+            Self::Off => "off",
```
spookyvision, line 209: you *could* use `#[derive(derive_more::Display)]`, though that won't give you all-lowercase

spookyvision (approved):

tools/x86_64-bootimager/src/lib.rs
hawkw: unfortunately, having it be all lower case is kinda load-bearing because the only reason this *has* a `Display` impl is so that `clap` can show the default value in the CLI help text, and it ought to be formatted the same way that `clap` will parse it there. I had _thought_ that deriving `clap::ValueEnum` would make that work without having to implement `Display` --- which the docs say it does --- but it didn't compile for me, so i just added this. whatever.

PR #338, opened 2024-10-03
Specify how to create a `Registration` in its doc comment.

BGR360:
As a brand new reader of the mnemos kernel docs, this was a bit difficult to discover. I started at `Registry::register`, and saw that it requires a `Registration`. But `Registration` has no constructor, and its docs didn't mention how one is created.


hawkw (approved):
Lovely, thanks for adding this!

PR #343, opened 2024-10-13
all platforms: update Mycelium dependencies

### LINKED ISSUE #337: mycelium-hal's interrupts trait is stupid
in ancient days, i was foolish and succumbed to hubris, by thinking that a cross-platform abstraction over interrupts was a cool idea and could be fun to do. so, the mycelium HAL uses a trait with a bunch of methods to represent generic interrupt handlers

it turns out that this sucks and is bad and i hate it. in particular it makes writing handlers for platform specific interrupts really painful, and makes it hard to e.g. use all the different timers that exist on x86. we should rip that thing out of the x86 MnemOS platform and actually put stuff in the IDT ourselves, like normal people.

hawkw:
This updates our dependencies on Mycelium crates to
hawkw/mycelium@ba56bb4d02f46fb59754b7b88bddc2e8ca99c1f5.

This picks up the new `maitake` timer API, as well as the upstream
change hawkw/mycelium#487 to allow raw access to the IDT in the Mycelium
x86_64 HAL, which we need to do an async UART driver (see #337).

While updating everything to use the new timer API, I also made some
tweaks to our existing timer code, including implementing 32-bit timer
rollover for the D1's TIMER0 (our current timestamp source on that
platform). This was an ancient TODO item since basically forever.

Fixes #337

### DIFF
--- a/platforms/allwinner-d1/d1-core/src/timer.rs
+++ b/platforms/allwinner-d1/d1-core/src/timer.rs
@@ -16,10 +16,12 @@ impl Timer0 {
 
         let _ = self.get_and_clear_interrupt();
 
+        // Start the timer counting down from 0xffff_ffff_ffff_ffff.
+
         // TODO(eliza): if it's zero, handle that lol --- when the IRQ fires
         // we need to increment some higher half counter and then reset the
         // timer to u32::MAX?
-        self.start_counter(0xFFFF_FFFF);
+        self.start_counter(u32::MAX);
 
         Clock::new(core::time::Duration::from_nanos(333), || {
             let timer0 = unsafe {
@@ -27,7 +29,10 @@ impl Timer0 {
                 // concurrently mutating the timer.
                 Self::steal()
             };
-            0xFFFF_FFFF - timer0.current_value() as u64
+            // Since timer 0 is counting *down*, we have to subtract its current
+            // value from the intial value to get an increasing timestamp for
+            // Maitake.
+            (u32::MAX - timer0.current_value()) as u64
         })
         .named("CLOCK_D1_TIMER0")
     }
--- a/platforms/allwinner-d1/d1-core/src/timer.rs
+++ b/platforms/allwinner-d1/d1-core/src/timer.rs
@@ -27,7 +27,7 @@ impl Timer0 {
                 // concurrently mutating the timer.
                 Self::steal()
             };
-            timer0.current_value() as u64
+            0xFFFF_FFFF - timer0.current_value() as u64
         })
         .named("CLOCK_D1_TIMER0")
     }
--- a/platforms/allwinner-d1/d1-core/src/timer.rs
+++ b/platforms/allwinner-d1/d1-core/src/timer.rs
@@ -2,10 +2,39 @@ pub use d1_pac::timer::tmr_ctrl::{
     TMR_CLK_PRES_A as TimerPrescaler, TMR_CLK_SRC_A as TimerSource, TMR_MODE_A as TimerMode,
 };
 use d1_pac::TIMER;
+use kernel::maitake::time::Clock;
 
 pub struct Timers {
     pub timer0: Timer0,
     pub timer1: Timer1,
+}
+
+impl Timer0 {
+    pub fn into_maitake_clock(mut self) -> Clock {
+        self.set_prescaler(TimerPrescaler::P8); // 24M / 8:  3.00M ticks/s
+        self.set_mode(TimerMode::PERIODIC);
+
+        let _ = self.get_and_clear_interrupt();
+
+        // TODO(eliza): if it's zero, handle that lol --- when the IRQ fires
+        // we need to increment some higher half counter and then reset the
+        // timer to u32::MAX?
+        self.start_counter(0xFFFF_FFFF);
+
+        Clock::new(core::time::Duration::from_nanos(333), || {
+            let timer0 = unsafe {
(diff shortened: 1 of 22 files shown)


platforms/allwinner-d1/src/lib.rs
```diff
@@ -344,22 +337,16 @@ impl D1 {
             }
         }
 
-        timer0.start_counter(0xFFFF_FFFF);
```
hawkw, line 347: I wonder if this line moving is why the D1 is wedged. The new code has moved the call to `start_counter` into `Timer0::into_maitake_clock`, which gets called *before* PLIC setup. Perhaps starting the counter before the PLIC is configured breaks something...

platforms/allwinner-d1/d1-core/src/timer.rs
```diff
@@ -2,12 +2,41 @@ pub use d1_pac::timer::tmr_ctrl::{
...
 
 pub struct Timers {
     pub timer0: Timer0,
     pub timer1: Timer1,
 }
 
+impl Timer0 {
+    pub fn into_maitake_clock(mut self) -> Clock {
+        self.set_prescaler(TimerPrescaler::P8); // 24M / 8:  3.00M ticks/s
+        self.set_mode(TimerMode::PERIODIC);
```
hawkw, line 15: Same prescaler and mode configuration we set in the older code, so that's presumably not the problem...

hawkw (commented):
trying to figure out why the timer is wedged on D1

source/kernel/src/test_util.rs
```diff
@@ -21,16 +21,25 @@ pub(crate) struct TestKernel {
...
         trace_init();
+
+        // TODO(eliza): this clock implementation is also used in Melpomene, so
+        // it would be nice if we could share it with melpo...
+        let clock = {
+            maitake::time::Clock::new(Duration::from_micros(1), || {
+                SystemTime::now()
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap()
+                    .as_micros() as u64
```
hawkw, line 32: probably we could make this use `Instant` rather than `SystemTime`, but we'd need to stuff an initial start time `Instant` in a global so that this function can get at it. eventually i'd like to redo the `Clock` API to let you actually put data into the `Clock` at init...

platforms/x86_64/core/src/interrupt.rs
```diff
@@ -128,9 +122,7 @@ impl hal_core::interrupt::Handlers<Registers> for InterruptHandlers {
     }
 
     fn timer_tick() {
-        if let Some(timer) = ptr::NonNull::new(TIMER.load(Ordering::Acquire)) {
-            unsafe { timer.as_ref() }.advance_ticks(1);
-        }
+        IDIOTIC_CLOCK_TICKS.fetch_add(1, Ordering::Relaxed);
```
hawkw, line 125: XXX ELIZA THIS SUCKS SO BAD LOL

platforms/pomelo/src/main.rs
```diff
@@ -65,14 +65,21 @@ async fn run_pomelo() {
...
-        timer_granularity: maitake::time::Duration::from_micros(1),
+    let settings = KernelSettings { max_drivers: 16 };
+
+    let clock = {
+        maitake::time::Clock::new(
+            // TODO(eliza): timer granularity chosen totally arbitrarily
+            Duration::from_micros(1),
+            || {
+                // TODO(anatol) please fix this for me thanks :)
+                unimplemented!("FIGURE OUT HOW TO GET A NORMAL TIMESTAMP OUT OF WASM LOL")
```
hawkw, line 76: one way to make this work with the current `Clock` interface is to stuff a `DateTime<Local>` into a global so it can be grabbed by this function. i'll probably just go do that. 

perhaps there's a nicer way to get a global integer timestamp on wasm, but i don't know what it is.

hawkw (commented):
more notes

hawkw:
oh, the D1 timer0 problem is actually something much stupider; it's just that the interval timer mode is counting down from `0xFFFF_FFFF` rather than up from 0, and `maitake`'s timer wheel expects timestamps to go up rather than down. So we need to change our timer impl to subtract the current value from `0xFFFF_FFFF`. lol.

hawkw:
> oh, the D1 timer0 problem is actually something much stupider; it's just that the interval timer mode is counting down from `0xFFFF_FFFF` rather than up from 0, and `maitake`'s timer wheel expects timestamps to go up rather than down. So we need to change our timer impl to subtract the current value from `0xFFFF_FFFF`. lol.

also, why did @jamesmunns make the downcounting interval start at `u16::MAX` rather than `u32::MAX`? IIUC we have an entire 32 bits of precision...doesn't this just eat half of our timestamp precision? or do i not actually get how the D1 timers work?

platforms/allwinner-d1/src/lib.rs
hawkw: UPDATE: lol it wasn't that, it was something way stupider: https://github.com/tosc-rs/mnemos/pull/343#issuecomment-2563944528

platforms/allwinner-d1/d1-core/src/timer.rs
hawkw: UPDATE: indeed it was not! https://github.com/tosc-rs/mnemos/pull/343#issuecomment-2563944528

